PR #30, opened 2024-06-02
Agora proposal handle

SolutionsEngineer:
The code will be edited to work with the API when the API is ready. 

https://github.com/idriss-crypto/browser-extensions/assets/106707979/1cb674fc-7894-4eec-8a11-d58589098091

https://github.com/idriss-crypto/browser-extensions/assets/106707979/edc88a14-19a3-4901-9657-cdfb981ed483


src/application/agora/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const TWITTER_HANDLE_TO_AGORA: Record<string, string> = {
+  optimism: 'optimism.eth',
+} as const;
```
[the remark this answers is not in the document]
nemmtor: actually maybe you can change it in both snapshot and agora? I think it's fine to just leave Record and remove `as const` since we don't need to know exact keys/values, it's just a lookup obj

src/application/agora/hooks/use-twitter-visible-agora-nodes.ts
```diff
@@ -0,0 +1,47 @@
+import { useMemo } from 'react';
+
+import { useTwitterUsersPooling } from 'src/host/twitter';
```
nemmtor, line 3: import should be without `src`, can you double check if there's any other place like that and fix it in scope of this PR please?

src/application/agora/hooks/index.ts
```diff
@@ -0,0 +1,2 @@
+export { useTwitterVisibleAgoraNodes } from './use-twitter-visible-agora-nodes';
```
nemmtor, line 1: is `useTwitterVisibleAgoraNodes` used outside of hooks directory? If not let's remove re-export

src/application/agora/schema.ts
```diff
@@ -0,0 +1,47 @@
...
+  proposalTemplate: z.object({
+    name: z.string(),
+    proposalTemplateId: z.number(),
+    votingStrategy: z.enum(['STANDARD']),
+    quorum: z.string(),
+    approvalThreshold: z.string(),
+    contractAddress: z.string(),
+    createBlock: z.string(),
+  }),
+});
```
nemmtor, line 34: I think we don't need most of this, let's cut down the schema to things we will use, if they change any field that we don't use we're still fine

src/application/agora/utils.ts
```diff
@@ -0,0 +1,67 @@
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
```
nemmtor, line 1: I think this eslint-disable is not needed here since we don't have gql string

src/application/agora/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+import { TwitterUserPoolingResult } from 'src/host/twitter';
```
nemmtor, line 2: please remove src

src/application/agora/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+export const getEndsInLabel = (daysUntil: number) => {
+  if (daysUntil < 0) {
+    return '-';
+  }
+
+  if (daysUntil === 0) {
+    return 'Ends today';
+  }
```
nemmtor, line 40: since these 2 functions are same in snapshot and agora I feel temptation to move it into some reusable function inside shared module, at the same time this might change in future so I am also fine to just duplicate this logic, WDYT?

src/host/twitter/hooks.ts
```diff
@@ -77,3 +83,13 @@ export const useTwitterExternalLinksPooling = () => {
...
+
+export const useTwitterLocationInfo = () => {
+  const location = useLocation();
+
+  const isTwitter = isTwitterHostname(location.hostname ?? '');
+  const isTwitterHandlePage = isTwitterHandlePathname(location.pathname ?? '');
+  const isTwitterHomePage = isTwitterHomePathname(location.pathname ?? '');
+
+  return { isTwitter, isTwitterHandlePage, isTwitterHomePage };
+};
```
nemmtor, line 95: can you reuse this in our other apps? There should be few todo's for that

src/application/agora/widgets/proposal-handle.container.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { ProposalData } from '../types';
+
+import { Proposal } from './proposal';
+
+interface Properties {
+  handle: string;
+}
+
+export const ProposalHandleContainer = ({ handle }: Properties) => {
+  const [currentProposalIndex, setCurrentProposalIndex] = useState<number>(0);
```
nemmtor, line 16: you don't need to pass generic here because typescript will infer the type based on provided default value

src/application/agora/widgets/proposal-handle.container.tsx
```diff
@@ -0,0 +1,75 @@
...
+      return previous + 1;
+    });
+  };
+
+  useEffect(() => {
+    const proposalResponse = proposalQuery.data?.proposals?.at(0);
+    if (proposalResponse) {
+      setCurrentProposal(proposalResponse);
+    }
+  }, [proposalQuery.data?.proposals]);
```
nemmtor, line 58: tanstack-query useQuery has extra option called `select`, by doing so you can adapt the data, I think you could do something like:
```ts
select: (response) => {{ ...response, data: { ...data, proposal: data.proposals[0] }}}
```

By doing so you can remove whole state for currentProposal and this useEffect and just use `proposalQuery.data.proposal`

src/application/agora/widgets/proposal-handle.container.tsx
```diff
@@ -0,0 +1,75 @@
...
+    return null;
+  }
+
+  return (
+    <Proposal
+      isPreviousProposalAvailable={isPreviousProposalAvailable}
+      isNextProposalAvailable={isNextProposalAvailable}
+      showNextProposal={showNextProposal}
+      showPreviousProposal={showPreviousProposal}
+      loadingNextProposal={proposalQuery.isLoading}
```
nemmtor, line 70: I suggest a few renames, usually when component is accepting some callback I name it with `on` prefix, like `onClick` etc. so by following this convention it could be:
```ts
onNext
onPrevious
```

also let's rename loading prop to `isLoading`

src/application/agora/widgets/proposal-main.container.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import { useTwitterVisibleAgoraNodes } from '../hooks';
+import { ProposalData } from '../types';
+import { GetAgoraProposalsCommand } from '../commands';
+
+import { Proposal } from './proposal';
+
+export const ProposalMainContainer = () => {
+  const [hiddenAgoraUsers, setHiddenAgoraUsers] = useState<string[]>([]);
+  const [currentProposalIndex, setCurrentProposalIndex] = useState<number>(0);
```
nemmtor, line 14: no need to pass number as generic here

src/application/agora/widgets/proposal-main.container.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return previous + 1;
+    });
+  };
+
+  useEffect(() => {
+    const proposalResponse = proposalQuery.data?.proposals?.at(0);
+    if (proposalResponse) {
+      setCurrentProposal(proposalResponse);
+    }
+  }, [proposalQuery.data?.proposals]);
```
nemmtor, line 65: same think about using `select` option as in the handle container

src/application/agora/widgets/proposal.tsx
```diff
@@ -0,0 +1,103 @@
...
+  onClose,
+  showPreviousProposal,
+  showNextProposal,
+}: Properties) => {
+  const proposalEndDateInMs = new Date(
+    data.proposalData.endTimestamp,
+  ).getTime();
+
+  const navigationButtonClassName =
+    'inline-flex h-7 select-none items-center justify-center rounded-md text-sm font-light text-[#444444] transition-all hover:text-[#7d848a] focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:pointer-events-none disabled:bg-white';
```
nemmtor, line 35: I avoid storing tailwind classNames in variables because it usually breaks tailwind ecosystem tools like autocomplete etc. I'd suggest to move the button itself to a separate component in `agora/components` and just inline the classname there in the html

src/application/agora/widgets/proposal.tsx
```diff
@@ -0,0 +1,103 @@
...
+            onClick={showNextProposal}
+            className={navigationButtonClassName}
+          >
+            Next
+          </Button>
+        </div>
+      </footer>
+      <div
+        className={`absolute top-0 h-1 ${loadingNextProposal ? 'left-0' : 'right-0'} ${loadingNextProposal ? 'w-full' : 'w-0'} animate-pulse rounded-full bg-gradient-to-r from-stone-300 via-stone-300 via-stone-400 via-stone-500 to-stone-300  delay-75 duration-200`}
+      />
```
nemmtor, line 100: this conditional classname inside template string won't work in tailwind, you should use `classes` function for that so:
```ts
classes('...', loadingNextProposal ? 'left-0 w-full' : 'right-0 w-0')
```

nemmtor: I am wondering if we need to set w-0, maybe we can just conditionally render whole div?

nemmtor (commented):
Nice work! Let's talk about pagination on discord

src/application/agora/widgets/proposal-main.container.tsx
nemmtor: actually it might be different when we go with that extra container as discussed on discord, I mean here you just need array of all proposals probably but still do `select` inside inner container

src/application/agora/app.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ErrorBoundary } from 'shared/monitoring';
+import { extractTwitterHandleFromPathname } from 'host/twitter';
+import { useTwitterLocationInfo } from 'host/twitter/hooks';
+
+import { ProposalHandleContainer, ProposalMainContainer } from './widgets';
+
+export const App = () => {
+  const { experimentalFeatures } = useExtensionSettings();
+  const { isTwitter, isTwitterHandlePage, isTwitterHomePage } =
+    useTwitterLocationInfo();
```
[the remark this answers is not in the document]
SolutionsEngineer: Thanks, glad to hear that!

src/application/agora/constants.ts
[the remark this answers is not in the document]
SolutionsEngineer: Yeah, that makes sense 😄

src/application/agora/hooks/use-twitter-visible-agora-nodes.ts
SolutionsEngineer: Sure, all occurrences updated :)

src/application/agora/hooks/index.ts
SolutionsEngineer: I've used it directly in the `ProposalMainContainer` component but forgot to remove the `useTwitterProposalsToDisplay` where it was used initially.

src/application/agora/schema.ts
SolutionsEngineer: Good idea

src/application/agora/utils.ts
SolutionsEngineer: Yeah, I was going to ask you about that. I'm not sure if the API will return only `ACTIVE` proposals, but maybe I will add handling past dates and move it to the shared module, I don't think that will cause any problems in the snapshot usage.

SolutionsEngineer: What do you think about JSDoc? I like using that in the common utils. to quickly understand the function, but I know some people don't like combining that with TS. 

image
image

src/host/twitter/hooks.ts
SolutionsEngineer: sure, I've found that in the `wallet-connect-modal.component.tsx` we also need to use that hook. Should we move it to the Shared module?

image

src/application/agora/widgets/proposal.tsx
SolutionsEngineer: I've added that w-0 -> w-full and duration-200 to animate the left-to-right transition. I felt that it makes the loading indicator more natural for the users.

src/application/agora/utils.ts
nemmtor: I like the idea about adding JSDoc! I was thinking about it but don't have experience with it. Feel free to add it wherever you see it's worth to do so.

I know that in JSDoc you can also type params/return type, should we still do it if we have ts types?

nemmtor: I will also start using JSDoc

nemmtor: > Yeah, I was going to ask you about that. I'm not sure if the API will return only `ACTIVE` proposals, but maybe I will add handling past dates and move it to the shared module, I don't think that will cause any problems in the snapshot usage.



> Yeah, I was going to ask you about that. I'm not sure if the API will return only `ACTIVE` proposals, but maybe I will add handling past dates and move it to the shared module, I don't think that will cause any problems in the snapshot usage.

Let's leave it as it is for now and see if api returns only active proposals and than we can adjust.

src/host/twitter/hooks.ts
nemmtor: @SolutionsEngineer let's not move it to shared, wallet-connect-modal case is very specific, we only check it because coinbase wallet is having problems on Twitter after newest update, ideally they fix it on their side and we don't need to worry about it. There should be eslint-disable, I've it on my fresh PR, not sure if it wasn't on main before and I just added it now.

image

If they don't fix it and we want to get rid of this problem that this place is breaking our rules we might decide to inject te list of banned wallets from higher layer which is allowed to use host module, something like this:
image

src/application/agora/widgets/proposal-handle.container.tsx
SolutionsEngineer: I've added the useState and useEffect to keep the value of the displayed proposal for the time of loading the next one. Is that achievable using the select param?

src/application/agora/widgets/proposal.tsx
```diff
@@ -0,0 +1,106 @@
+import { classes } from 'shared/ui/utils';
+import { Chip, WidgetBase } from 'shared/ui/components';
+import {
+  getDifferenceInDays,
+  getEndsInLabel,
+} from 'shared/ui/utils/date-utils';
```
nemmtor, line 6: can you export it from shared/ui/utils ?

src/application/agora/widgets/proposal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { ProposalData } from '../types';
+import { getProposalUrl } from '../utils';
+import { NavigationButton } from '../components/navigation-button';
```
nemmtor, line 10: it won't change anything but just to follow convention please make index.ts to re-export components

src/application/polymarket/hooks/use-warpcast-markets.ts
```diff
@@ -1,8 +1,7 @@
 import { useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 
-import { useWarpcastExternalLinksPooling } from 'src/host/warpcast';
-
+import { useWarpcastExternalLinksPooling } from '../../../host/warpcast';
```
nemmtor, line 4: for imports across different layers like application->host we should use absolute imports, `host/warpcast` should work

src/application/snapshot/hooks/use-twitter-visible-snapshots.ts
```diff
@@ -1,7 +1,6 @@
 import { useMemo } from 'react';
 
-import { useTwitterUsersPooling } from 'src/host/twitter';
-
+import { useTwitterUsersPooling } from '../../../host/twitter';
```
nemmtor, line 3: same as above

src/application/snapshot/utils.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/restrict-template-expressions */
-import { TwitterUserPoolingResult } from 'src/host/twitter';
+import { TwitterUserPoolingResult } from '../../host/twitter';
```
nemmtor, line 2: same as above

src/application/snapshot/widgets/proposal.tsx
```diff
@@ -1,14 +1,12 @@
...
-import { ProposalData } from '../types';
 import {
-  getDaysUntil,
+  getDifferenceInDays,
   getEndsInLabel,
-  getProposalAuthor,
-  getProposalUrl,
-  getUserUrl,
-} from '../utils';
+} from 'shared/ui/utils/date-utils';
```
nemmtor, line 6: please re-export from shared/ui/utils

PR #31, opened 2024-06-20
Tally proposal handle

SolutionsEngineer:
- tally feature
image

- minor style changes to the agora feature
image
 
- minor style changes to snpshot feature
image

- service status call API path change
https://www.idriss.xyz/service-status -> https://api.idriss.xyz/service-status

- replaced hardcoded dao handles with the ones fetched from DB
The Dao handles are now fetched from https://api.idriss.xyz/dao-twitter-handles


tailwind.config.mjs
```diff
@@ -12,12 +12,39 @@ export default {
...
+            100: '#FFE6E7'
+          },
+          border: {
+            primary: '#EAECF0',
+            gray: { 700: '#2D3748' },
+          },
+          text: {
+            primary: '#1D2939'
+          }
+        },
```
nemmtor, line 47: thanks for putting these in tailwind config!

src/shared/extension/commands/get-dao-handles.ts
```diff
@@ -0,0 +1,44 @@
...
+type Payload = Record<string, never>;
+
+const responseSchema = z.record(z.string(), z.record(z.string(), z.string()));
+type Response = z.infer<typeof responseSchema>;
+
+export class GetDaoHandlesCommand extends Command<Payload, Response> {
+  public readonly name = 'GetDaoHandlesCommand' as const;
+
+  constructor(
+    public payload: Record<string, never>,
```
nemmtor, line 19: you can type it with already existing `Payload` type

src/shared/extension/hooks/use-dao-handles.ts
```diff
@@ -0,0 +1,13 @@
+import { useCommandQuery } from 'shared/messaging';
+
+import { GetDaoHandlesCommand } from '../commands';
+
+export const useDaoHandles = (application: string) => {
+  return useCommandQuery({
+    staleTime: Number.POSITIVE_INFINITY,
```
nemmtor, line 7: since default staleTime is now also infinity you can remove this1

src/shared/extension/hooks/use-dao-handles.ts
```diff
@@ -0,0 +1,13 @@
+import { useCommandQuery } from 'shared/messaging';
+
+import { GetDaoHandlesCommand } from '../commands';
+
+export const useDaoHandles = (application: string) => {
```
nemmtor, line 5: I think this1 is mapping from twitter to username on given application, right?
If yes than we should probably move this1 to `host/twitter` and maybe we can try to apply more meaningfull name like `useHandleToUsernameMap`

nemmtor: Also maybe it's worth to change api endpoint url to make it more obvious that it's specific for twitter CC @lennardevertz

src/application/snapshot/hooks/use-twitter-visible-snapshots.ts
```diff
@@ -10,10 +11,11 @@ interface Properties {
 
 export const useTwitterVisibleSnapshots = ({ hidden }: Properties) => {
   const { tweetAuthors } = useTwitterScraping();
+  const { data: daoHandles } = useDaoHandles('snapshot');
 
   const snapshotUserNodes = useMemo(() => {
-    return getSnapshotUsernameNodes(tweetAuthors);
-  }, [tweetAuthors]);
+    return getSnapshotUsernameNodes(daoHandles ?? {}, tweetAuthors);
+  }, [daoHandles, tweetAuthors]);
```
nemmtor, line 18: I am thinking that this function `getSnapshotUsernameNodes` is unnecessary now, we need to pass tweetAuthors and daoHandles just to iterate inside and apply another fn. I also have similar feeling to `getSnapshotFromTwitterUsername`, it made more sense when we had constant map but now we are getting that data from api call so maybe we can simplify and do it directly here, something like:
```tsx
  const snapshotUserNodes = useMemo(() => {
    return tweetAuthors.map((author)=>{
      const snapshotName = daoHandles[handle.toLowerCase()];
      return snapshotName ? { snapshotName, ...result} : undefined;
    }).filter(Boolean)
  }, [daoHandles, tweetAuthors]);
```

and you could remove 2 functions mentioned above

src/application/snapshot/widgets/proposal-handle.container.tsx
```diff
@@ -10,7 +11,8 @@ interface Properties {
 }
 
 export const ProposalHandleContainer = ({ handle }: Properties) => {
-  const snapshotName = getSnapshotFromTwitterUsername(handle);
+  const { data: daoHandles } = useDaoHandles('snapshot');
+  const snapshotName = getSnapshotFromTwitterUsername(daoHandles ?? {}, handle);
```
nemmtor, line 15: similar to `hooks/use-twitter-visible-snapshots.ts` - you could probably do:
```tsx
  const { data: daoHandles } = useDaoHandles('snapshot');
  const snapshotName = daoHandles ? daoHandles[handle] : undefined;
```

src/application/tally/widgets/proposal.tsx
```diff
@@ -0,0 +1,136 @@
...
+              proposalDetails.creator.ens.length > 0 ||
+              proposalDetails.creator.address.length > 0
+                ? getOrganizationDelegateUrl(
+                    proposalDetails.organization.slug ?? '',
+                    proposalDetails.creator.ens.length > 0
+                      ? proposalDetails.creator.ens
+                      : proposalDetails.creator.address,
+                  )
+                : getOrganizationUrl(proposalDetails.organization.slug ?? '')
+            }
```
nemmtor, line 60: maybe we can hide this conditional logic behind a function `getAuthorUrl` ? This function would receive whole `proposalDetails` obj as argument and return url

src/application/tally/widgets/proposal.tsx
```diff
@@ -0,0 +1,136 @@
...
+            className="font-semibold font-semibold hover:underline"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            {proposalDetails.creator.name.length > 0
+              ? proposalDetails.creator.name
+              : proposalDetails.creator.address}
+          </a>
```
nemmtor, line 68: I also suggest to hide this1 behind fn `getProposalAuthorLabel`

src/application/tally/widgets/proposal.tsx
```diff
@@ -0,0 +1,136 @@
...
+              disabled={!isPreviousProposalAvailable}
+              onClick={onPrevious}
+              iconProps={{ name: 'ArrowLeftIcon' }}
+              className="px-0"
+            />
+            <IconButton
+              disabled={!isNextProposalAvailable}
+              onClick={onNext}
+              iconProps={{ name: 'ArrowRightIcon' }}
+              className="px-0"
```
nemmtor, line 123: isn't this px-0 gonna be problematic in mobile devices because of small touch area?

src/application/tally/widgets/proposal.tsx
```diff
@@ -0,0 +1,136 @@
...
+            />
+          </div>
+        )}
+      </footer>
+      <div
+        className={classes(
+          'absolute top-0 h-1 animate-pulse rounded-full bg-gradient-to-r from-stone-300 via-stone-300 via-stone-400 via-stone-500 to-stone-300  delay-75 duration-200',
```
nemmtor, line 130: `'via-stone-300' applies the same CSS properties as 'via-stone-400' and 'via-stone-500'.(cssConflict)`

src/application/tally/widgets/proposal.tsx
```diff
@@ -0,0 +1,136 @@
...
+          >
+            <Chip
+              className=" mr-2 inline-flex h-7 select-none items-center justify-center rounded-md border border-tally-gray-700 bg-tally-gray-700 px-4 py-2 text-sm font-semibold leading-6 text-white transition-all duration-200 ease-in-out hover:bg-tally-gray-900 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50"
+              variant="info"
+              width="long"
+            >
+              {proposalDetails.status.toLowerCase() === 'active'
+                ? 'Vote'
+                : 'View'}
+            </Chip>
```
nemmtor, line 108: do we know finite list of proposal `status`? we could maybe narrow the type and make a constant to map status to label and apply fn on top of it:
```ts
const STATUS_TO_LABEL: Record<Status, string> = {
  active: 'active',
  foo: 'view',  
}

const getStatusLabel = (status: Status) => {
  return STATUS_TO_LABEL[status];
}
```

src/application/tally/widgets/organization-proposals.container.tsx
```diff
@@ -0,0 +1,142 @@
...
+  } = useTallyContext();
+
+  const [fetchedProposals, setFetchedProposals] = useState<ProposalData[]>([]);
+  const [hasMoreProposalsToFetch, setHasMoreProposalsToFetch] =
+    useState<boolean>(true);
+
+  const organizationInfoQuery = useCommandQuery({
+    command: new GetOrganizationInfoCommand({ tallyName: tallyName ?? '' }),
+    enabled: tallyName ? tallyName.length > 0 : false,
+    staleTime: Number.POSITIVE_INFINITY,
```
nemmtor, line 41: you can remove staleTime

src/application/tally/tally.context.tsx
```diff
@@ -0,0 +1,106 @@
...
+    new Map<string, OrganizationInfo>(),
+  );
+
+  const getOrganizationInfo = (tallyName: string) => {
+    let organization = organizationsInfo.get(tallyName);
+    if (!organization) {
+      organization = createNewOrganizationsInfoMap();
+      setOrganizationsInfo(
+        new Map(organizationsInfo).set(tallyName, organization),
+      );
```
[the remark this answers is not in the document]
nemmtor: if new proposition of pagination makes sense than it's probably not needed anymore anyway

src/application/tally/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      return {
+        tallyName,
+        ...result,
+      };
+    })
+    .filter(Boolean);
+};
+
+export const getStatusBadgeColorClassNames = (status: string) => {
```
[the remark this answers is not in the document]
nemmtor: image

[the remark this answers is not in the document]
nemmtor: on the other hand I see your intention - we just care about 2 statuses and rest is just "other" so IMHO it's fine to have it as a string

src/application/tally/commands/get-proposals.command.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      const json = await response.json();
+
+      const validResponse = getProposalsResponseSchema.parse(json);
+      const activeProposals = {
+        nodes: validResponse.data.proposalsV2.nodes.filter((proposal) => {
+          return proposal.status === 'active';
+        }),
+        pageInfo: validResponse.data.proposalsV2.pageInfo,
+      };
```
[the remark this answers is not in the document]
nemmtor: ah I remember, they don't 🫠

src/application/tally/widgets/proposal.tsx
nemmtor: I posted statuses image above and also mentioned it's fine to have status as string so here we could just make a fn to return active if status is active or view if other

nemmtor:
🚀 nice, let's also try to simplify pagination to get rid of context

src/shared/extension/hooks/use-dao-handles.ts
lennardevertz: I can change it to `/dao-twitter-handles` for example. Is that better?

src/application/tally/commands/get-proposals.command.ts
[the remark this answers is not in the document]
SolutionsEngineer: Yeah, unfortunatelly 🥲

src/application/tally/commands/get-proposals.command.ts
```diff
@@ -0,0 +1,82 @@
...
+                limit: 1,
+                afterCursor: this.payload.afterCursor,
+              },
+            },
+          },
+        }),
+        headers: {
+          'Content-Type': 'application/json',
+          'Api-Key':
+            'a0a4cd00bb6953720c9c201c010cdd36a563e65c97e926a36a8acdfcd1d1eeb7',
```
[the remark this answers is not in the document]
SolutionsEngineer: This is an API key from one of their public repos, so we don't need to worry about exposing it. But yeah, proxying the API will be necessary

src/application/tally/hooks/use-twitter-visible-tally-users-nodes.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const visibleTallyNodes = useMemo(() => {
+    return tallyUserNodes
+      .filter((tallyNode) => {
+        return !hidden.includes(tallyNode.tallyName);
+      })
+      .map((tallyNode) => {
+        return {
+          name: tallyNode.tallyName,
+          top: tallyNode.top ?? 0,
```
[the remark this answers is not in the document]
SolutionsEngineer: You're right. It was just repalcing 0 with 0 😅

src/application/tally/widgets/organization-proposals.container.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+  const proposalQuery = useCommandQuery({
+    command: new GetTallyProposalsCommand({
+      tallyUserId: organizationInfoQuery.data?.id?.toString() ?? '',
+      afterCursor: currentProposal?.id ? `0;${currentProposal.id}` : null,
+    }),
+    enabled: isProposalQueryEnabled,
+    retry: 5,
+    retryDelay: 1800,
+    staleTime: Number.POSITIVE_INFINITY,
```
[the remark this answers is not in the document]
SolutionsEngineer: I don't know if it's a matter of some unmerged changes but without the staleTime it looks like the request are not cached.

src/shared/extension/hooks/use-dao-handles.ts
SolutionsEngineer: I don't know if it's a matter of some unmerged changes but without the staleTime it looks like the request are not cached.

src/shared/extension/hooks/use-dao-handles.ts
SolutionsEngineer: updated :)

src/application/tally/widgets/proposal.tsx
SolutionsEngineer: changed to px-1

nemmtor (approved):

PR #35, opened 2024-06-29
Snapshot pagination

SolutionsEngineer:
Added pagination to the snapshot widget

https://github.com/idriss-crypto/browser-extensions/assets/106707979/254e746c-4c43-490e-b96f-e3885c55a3f1


src/application/snapshot/commands/get-proposal.ts
```diff
@@ -51,18 +59,24 @@ export class GetProposalCommand extends Command<Payload, ProposalData> {
...
       return new OkResult({
-        ...proposal,
-        author: {
-          address: proposal.author,
-          resolvedAddress,
-        },
-      });
+        proposal: proposalWithResolvedAddress,
+        hasNextProposal: !!validationResult.data.data.proposals[1],
+      } as Response);
```
nemmtor, line 79: Maybe we could simplify it a little bit:
```ts
const [currentProposal, nextProposal] = validationResult.data.data.proposals;
const hasNextProposal = Boolean(nextProposal);
if(!current){
  return new OkResult({
    proposal: null,
    hasNextProposal: hasNextProposal, // probably should be hardcoded to false?
  })
}

const resolvedAddress = await resolveAddress(proposal.author);
return new OkResult({
    proposal: {
      ...proposal,
      author: { ...proposal.author, resolvedAddress } },
      hasNextProposal
  })
```

src/application/snapshot/utils.ts
```diff
@@ -1,36 +1,28 @@
...
+        title
+        body
+        end
+        author
+        space {
+          id
+        }
+      }
+    }`;
 };
```
nemmtor, line 26: after refactoring we only pass a single snapshotName, can you change argument type to `String!` and rename to `$snapshotName` to highlight that it should be a single snapshot passed?

nemmtor: btw. that's interesting that it was fine for GQL that you were passing a string instead of array of strings

src/application/snapshot/widgets/proposal.tsx
```diff
@@ -8,19 +15,29 @@ interface Properties {
...
+  top,
+  pagination,
+  isLoading,
+  onClose,
+}: Properties) => {
   return (
     <WidgetBase
       className={classes(
-        'rounded-lg bg-[#2d2d2d] text-xs leading-tight',
+        'rounded-[0.375rem] bg-[#2d2d2d] text-xs leading-tight',
```
nemmtor, line 34: can be simplified to `rounded-md`

src/application/snapshot/app.tsx
```diff
@@ -14,14 +17,23 @@ export const App = () => {
     twitterHandleFromPathname,
   } = useTwitterLocationInfo();
 
+  const { data: daoHandles } = useHandleToUsernameMap('snapshot');
+  const snapshotHandle =
+    isTwitterHandlePage && daoHandles
+      ? daoHandles[twitterHandleFromPathname.toLowerCase()]
+      : undefined;
+
```
nemmtor, line 25: Minor but could be simplified to
```ts
  const snapshotHandle = isTwitterHandlePage
    ? daoHandles?.[twitterHandleFromPathname.toLowerCase()]
    : undefined;
```

src/application/snapshot/widgets/organization-proposals.container.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const showPreviousProposal = () => {
+    if (!isPreviousProposalAvailable || isLoadingProposal) {
+      return;
+    }
+
+    setCurrentProposalIndex((previous) => {
+      return previous - 1;
+    });
+  };
```
nemmtor, line 54: maybe we should disable previousProposal while isLoading - same as you did for isNextProposalAvailable and later inside `showPreviousProposal` you can only check if it's available without need to check for loading state

src/application/snapshot/widgets/organization-proposals.container.tsx
```diff
@@ -0,0 +1,87 @@
...
+      snapshotName: snapshotHandle ?? '',
+      pageNumber: currentProposalIndex,
+    }),
+    enabled: snapshotHandle ? snapshotHandle.length > 0 : false,
+    staleTime: Number.POSITIVE_INFINITY,
+    placeholderData: (previousData) => {
+      return previousData;
+    },
+  });
+
```
nemmtor, line 36: I am wondering if it's still needed to check for snapshotHandle length etc. in typescript it seems like it's always a string, can you double check that please?

src/application/snapshot/commands/get-proposal.ts
SolutionsEngineer: Nice, I like this approach 👌

src/application/snapshot/utils.ts
SolutionsEngineer: Yeah, quite weird, but actuially we need to pass [String!], otherwise we are getting error: 
image

To make it consistent I've changed the payload type
image

src/application/snapshot/widgets/organization-proposals.container.tsx
SolutionsEngineer: oh yeah, but as I'm thinking about it, checking isLoading as a condition to show previous proposals doesn't make sense as we cache them. I've removed the isLoadingProposal from that condition. Let me know what you think about that.

src/application/snapshot/widgets/organization-proposals.container.tsx
SolutionsEngineer: You are right, we don't need that anymore

src/application/snapshot/utils.ts
nemmtor: ok thanks 😄

src/application/snapshot/widgets/organization-proposals.container.tsx
nemmtor: makes sense, thanks 👍🏻

src/application/snapshot/widgets/organization-proposals.container.tsx
```diff
@@ -24,10 +24,9 @@ export const OrganizationProposalsContainer = ({
 
   const proposalQuery = useCommandQuery({
     command: new GetProposalCommand({
-      snapshotName: snapshotHandle ?? '',
+      snapshotNames: [snapshotHandle] ?? [],
```
nemmtor, line 27: do we need this fallback?

nemmtor (approved):

src/application/snapshot/widgets/organization-proposals.container.tsx
SolutionsEngineer: not at all :D good catch!

nemmtor (approved):

PR #36, opened 2024-07-01
Widget tabs

SolutionsEngineer:


src/shared/ui/providers/widget-tabs.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export const useWidgetTabs = () => {
+  const context = useContext(WidgetTabsContext);
+  if (!context) {
+    throw new Error('useWidgetTabs must be used within a WidgetTabsProvider');
+  }
+
+  return context;
+};
```
nemmtor, line 106: we have abstraction for that:
```tsx
export const useWidgetTabs = createContextHook(WidgetTabsContext);  
```

src/shared/ui/components/widget-tab/widget-tab.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+  useEffect(() => {
+    if (userTabs === null && onClose) {
+      onClose();
+    }
+  }, [onClose, userTabs]);
+
+  if (!userTabs || userTabs.length === 0) {
+    return;
+  }
+
```
nemmtor, line 57: hmm I am surprised that typescript doesn't complain here but usually returning null is expected :think:

src/shared/ui/components/widget-tab/widget-tab.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+  useEffect(() => {
+    addWidgetTab(twitterHandle, tabName);
+  }, [addWidgetTab, tabName, twitterHandle]);
+
+  useEffect(() => {
+    if (userTabs === null && onClose) {
+      onClose();
+    }
+  }, [onClose, userTabs]);
```
nemmtor, line 52: I am wondering if these useEffect could be removed and this could be handled by doing:
```tsx
      onClose={() => {
        removeAllUserWidgets(twitterHandle);
        onClose?.();
      }}
```

WDYT?

src/shared/ui/components/widget-tab/widget-tab.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+      className={classes(
+        'z-10 overflow-visible rounded-[0.375rem] text-xs leading-tight ',
+        className,
+        {
+          'z-20': userPreferredTab === tabName,
+          'rounded-tl-none': userTabs.length > 1,
+          'bg-[#2d2d2d]': theme === 'dark',
+          'bg-white': theme === 'bright',
+        },
+      )}
```
nemmtor, line 69: I am wondering if className shouldn't be the last one because it's something passed from higher layer and higher layer may always want to override some styles, in tailwind and twMerge util (which we use under classess abstraction) latest class will be applied, so imagine you want in some specific case override background to `bg-red` it still won't be applied because later `bg-white` is applied, should we move className as last arg?

src/shared/ui/components/widget-tab/widget-tab.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+    if (index !== -1) {
+      return index * 100;
+    }
+
+    return 0;
+  }, [tabName, userTabs]);
+
+  useEffect(() => {
+    addWidgetTab(twitterHandle, tabName);
+  }, [addWidgetTab, tabName, twitterHandle]);
```
nemmtor, line 46: is it okay if this effect will re-run multiple times in a lifecycle of single WidgetTab? I am wondering if it would be better if we used `useEffectOnce` coming from `react-use` library which we have installed, but I might be completely wrong :think:

src/shared/ui/components/widget-tab/tab-handle.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+  isActive,
+  theme,
+  onClick,
+}: TabHandleProperties) => {
+  return (
+    <div
+      onClick={onClick}
+      style={{
+        left,
+      }}
```
nemmtor, line 18: Currently we calculate `left` property based on index, I was thinking if this would be possible by having a single div wrapping all `TabHandles`, this div would have absolute position, same as this TabHandle and it would just display all Tabs inside of it, by doing so you don't need to calculate `left` property as they will just be aligned in a single row, probably with addition of `display: flex`, etc. WDYT?

src/shared/ui/components/widget-tab/tab-handle.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+export const TabHandle = ({
+  left,
+  imageSrc,
+  name,
+  isActive,
+  theme,
+  onClick,
+}: TabHandleProperties) => {
+  return (
+    <div
```
nemmtor, line 14: can we use button since it's a clickable element?

src/shared/ui/components/widget-tab/tab-handle.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+      className={classes(
+        'absolute -top-[25px] flex h-[25px] w-[100px] cursor-pointer flex-row items-center gap-[6px] rounded-t-lg pl-2 pt-1 font-bold transition-all duration-75 ease-linear',
+        {
+          'bg-[#2d2d2d]': theme === 'dark',
+          'bg-white': theme === 'bright',
+          '-top-[23px] -translate-x-[3px] scale-95 brightness-75': !isActive,
+        },
+      )}
+    >
+      {imageSrc && <img className="h-3/4" src={imageSrc} />}
```
nemmtor, line 28: I am thinking if it would be better to hardcode height in px values, not sure.
I am imaging a situation that tab will grow in height one day and should image grow too?

I am actually find with both, so it's up to you :)

src/application/tally/widgets/organization-proposals.container.tsx
```diff
@@ -32,7 +32,7 @@ export const OrganizationProposalsContainer = ({
 
   const proposalQuery = useCommandQuery({
     command: new GetTallyProposalsCommand({
-      twitterHandle: tallyName ?? '',
+      twitterHandle: twitterHandle ?? '',
```
nemmtor, line 35: probably fallback is not needed anymore

src/application/tally/widgets/organization-proposals.container.tsx
```diff
@@ -97,12 +97,13 @@ export const OrganizationProposalsContainer = ({
     onPrevious: showPreviousProposal,
     onNext: showNextProposal,
   };
-  if (!currentProposal || !tallyName) {
+  if (!currentProposal || !twitterHandle) {
```
nemmtor, line 100: also this check for twitterHandle is not needed (looking at types it's always string)

src/shared/ui/components/widget-tab/widget-tab.component.tsx
SolutionsEngineer: That is a hacky way to close all widgets. Clicking on the `x` calls `removeAllUserWidgets`, which sets `userTabs` to `null`. This triggers a `useEffect`, causing other widgets to call their own `onClose` functions.

src/shared/ui/components/widget-tab/widget-tab.component.tsx
SolutionsEngineer: Agree! Good point

src/shared/ui/components/widget-tab/widget-tab.component.tsx
SolutionsEngineer: oh, had no idea about that function. In theory, the dep params that are there should not change, but I'll use the `useEffectOnce`

src/shared/ui/components/widget-tab/tab-handle.component.tsx
SolutionsEngineer: I was also considering that, but I wasn't sure where to place the wrapper since we don't have a higher-level component that groups all the widgets together

src/shared/ui/components/widget-tab/tab-handle.component.tsx
SolutionsEngineer: updated, fixed height seems to be safer :D

nemmtor (approved):

PR #40, opened 2024-07-27
feat: gitcoin v2

nemmtor:
- bring back gitcoin widget back to life
- add higher layer so GitcoinDonationWidget and IDrissSendWidget can be composed on that higher layer and we can control which one to inject, we will use same higher layer to refactor WidgetTabs. Note that for now I named this `final` but I plan to apply better name in future (I didn't do it now because of possible merge conflicts and huge diff), probably `application` will be renamed to `widgets` and `final` will be renamed to `application`. Note that this new higher layer removed the need to have `container` components inside current `application` layer, I think it adds a benefit of cleaner code also.
- added script for unused exports and cleaned all of the unused ones from code
- change the way we fetch gitcoin rounds (via IDriss proxy) and add schema validation to the response
- fix few wrong imports


.storybook/tailwind.css
```diff
@@ -760,6 +754,9 @@ video {
 .w-full {
   width: 100%;
 }
+.min-w-\[84px\] {
```
SolutionsEngineer, line 757: Do we need to add the arbitrary values here?

SolutionsEngineer: nvm, forgot that it's autogenerated😅

src/application/gitcoin/hooks/use-recipients.ts
```diff
@@ -0,0 +1,73 @@
...
+            user.value.toLowerCase()
+          );
+        });
+
+        if (!application) {
+          return;
+        }
+
+        const { value, node, top } = user;
+        const username = value;
```
SolutionsEngineer, line 46: what do you think about changing this to `const { value: username, node, top } = user;` and removing the line below (`const username = value;`)?

src/application/gitcoin/utils.ts
```diff
@@ -18,6 +15,7 @@ export const selectTwitterApplications = (
   return twitterApplications;
 };
 
+// TODO: consider using lodash.groupBy
```
SolutionsEngineer, line 18: Not sure if that will work in this case, but there is also a new Object groupBy feature available.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy#examples

src/application/gitcoin/widgets/donation-widget.tsx
```diff
@@ -2,203 +2,232 @@ import { memo, useCallback, useState } from 'react';
...
         return (
           <>
-            {feeInDollars > 0 ? (
-              <p className="flex w-full items-center justify-end space-x-1 text-xs text-gray-500">
-                <span>{`$${feeInDollars}`}</span>
-                <span>+</span>
-                <GasIcon />
-              </p>
-            ) : null}
+            <IdrissSend.Heading>Donate to @${username}</IdrissSend.Heading>
```
SolutionsEngineer, line 164: is the `$` sign needed here?

SolutionsEngineer (approved):
🚀

src/application/gitcoin/utils.ts
nemmtor: oh nice1, I didn't know this1! I think we will need to change lib in tsconfig because right now typescript doesn't see this method. I updated TODO for now, is it okay to leave it like that?

src/application/gitcoin/widgets/donation-widget.tsx
nemmtor: no, good catch!

PR #41, opened 2024-07-28
feat: gitcoin v2

nemmtor:


src/host/twitter/utils.ts
```diff
@@ -15,3 +15,17 @@ export const extractTwitterHandleFromPathname = (pathname: string) => {
 export const isTwitterHomePathname = (pathname: string) => {
   return pathname.startsWith('/home');
 };
+
+export const isHandleNode = (element: HTMLElement) => {
+  let currentElement: HTMLElement | null = element;
+  for (let index = 0; index < 10; index++) {
+    currentElement = currentElement.parentElement;
+    if (!currentElement) {
+      return false; // If there are less than 3 parents
```
SolutionsEngineer, line 24: I'm unsure if I understand the comment here, should there be `10` instead of `3`?

src/application/gitcoin/hooks/use-recipients.ts
```diff
@@ -0,0 +1,72 @@
...
+          return;
+        }
+
+        const { value: username, node, top } = user;
+
+        const isHandleUser =
+          handle === username && isHandleNode(node as HTMLElement);
+
+        const nodeToInject: HTMLElement | null = isHandleUser
+          ? node.querySelector('div > div > div > span') ??
```
SolutionsEngineer, line 51: What do you think about moving the paths to shared constants?

SolutionsEngineer (dismissed):

src/host/twitter/utils.ts
nemmtor: yeah correct :D

nemmtor: instead of fixing comment I moved this magic number to some meaningful constant

src/application/gitcoin/hooks/use-recipients.ts
nemmtor: good idea, I moved it to host/twitter/utils, it would be hard to find some meaningful name for constants for this :D

lennardevertz (approved):

PR #47, opened 2024-08-01
agora api adjustments

SolutionsEngineer:
https://github.com/user-attachments/assets/c524ab01-7598-480a-96fc-419e8ee2f0fb

### DIFF
--- a/src/application/agora/commands/get-agora-proposals.ts
+++ b/src/application/agora/commands/get-agora-proposals.ts
@@ -5,19 +5,22 @@ import {
   OkResult,
 } from 'shared/messaging';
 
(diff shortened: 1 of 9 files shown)


lennardevertz:
Looks good to me!
Maybe we can review the definition of `AGORA_ETHERSCAN_WEBSITE_URL`, as it is simply the block explorer of Optimism. It can be taken from `CHAIN` in `src/shared/web3/constants.ts`

SolutionsEngineer:
Thanks @lennardevertz, I've removed the `AGORA_ETHERSCAN_WEBSITE_URL` const and used the `CHAIN`

nemmtor (approved):

PR #52, opened 2024-08-13
feat: refactor and migrate proposals widgets to warpcast and supercast

nemmtor:


src/application/agora/widgets/proposal-main.container.tsx
```diff
@@ -1,40 +1,21 @@
...
-      return previousData;
-    },
-  });
+const Base = ({ className, onClose }: Properties) => {
+  const [currentProposalIndex, setCurrentProposalIndex] = useState(0);
 
+  const proposalQuery = useProposalsQuery({ offset: currentProposalIndex });
   const currentProposal = proposalQuery.data?.proposal;
   const isLoadingProposal =
     proposalQuery.isLoading || proposalQuery.isPlaceholderData;
```
SolutionsEngineer, line 21: I think we can get rid of isPlaceholder, since query don't use the placeholderData

src/application/snapshot/widgets/organization-proposals.container.tsx
```diff
@@ -1,40 +1,25 @@
...
-      return previousData;
-    },
+  const proposalQuery = useProposalsQuery({
+    officialName,
+    pageNumber: currentProposalIndex,
   });
-
   const currentProposal = proposalQuery.data?.proposal;
 
   const isLoadingProposal =
```
SolutionsEngineer, line 25: I think the isLoadingProposal can be changed to proposalQuery.isLoading, as we are not using placeholderData

src/application/tally/widgets/organization-proposals.container.tsx
```diff
@@ -30,17 +24,9 @@ export const OrganizationProposalsContainer = ({
...
-    }),
-    retry: 5,
-    retryDelay: 1800,
-    staleTime: Number.POSITIVE_INFINITY,
-    placeholderData: (previousData) => {
-      return previousData;
-    },
+  const proposalQuery = useProposalsQuery({
+    username,
+    afterCursor: currentProposalCursor,
```
SolutionsEngineer, line 29: as above, the isPlaceholderData check can be removed from the isLoadingProposal variable

src/final/hooks/use-user-widgets.ts
```diff
@@ -0,0 +1,79 @@
...
+        (recipient) => {
+          return recipient.username === username;
+        },
+      );
+
+      if (gitcoinRecipientsForThisUsername.length > 0) {
+        return gitcoinRecipientsForThisUsername;
+      }
+
+      const idrissRecipientsForThisUsername = idrissSendWidgets.filter(
```
SolutionsEngineer, line 61: Just to confirm that I got the logic right, we use idrissRecipients only as a fallback if no gitcoinRecipients are found, right?

nemmtor: yes, exactly, gitcoin is bigger priority, if there's no active gitcoin round we use our send widget

SolutionsEngineer (approved):
Awesome job! Adding that higher layer to manage the widgets will definitely pay off in the future 🙌🙌
TotallyawesomeTotallyAwesomeGIF

PR #57, opened 2024-08-27
added scroll to the select component

SolutionsEngineer:
https://github.com/user-attachments/assets/64742bc1-e111-4e9d-8e23-e6c147df7bb4

https://github.com/user-attachments/assets/c57dded5-d59a-45b3-86e3-6ebded65be91


tailwind.css
```diff
@@ -1,3 +1,9 @@
 @tailwind base;
+
+* {
+  scrollbar-width: thin;
+  scrollbar-color: gray rgb(239, 239, 239);
+}
+
```
nemmtor, line 7: I am not sure if this will work because tailwind.css is generated and due to some limitations and quick decisions is also living in multiple files, please check these commands in package.json: `tailwind:react` and `tailwind:legacy`

I think we could just move these classNames into this select.component, WDYT?
Btw. this is not built in tailwind but you can use custom classNames with this:
```ts
[scrollbar-color:gray_rgb(239,_239,_239)] [scrollbar-width:thin]
```

I'd suggest to stick to hex colors tho:
```ts
[scrollbar-color:gray_#efefef] [scrollbar-width:thin]
```

I hope it works!

nemmtor (commented):
Awesome stuff, I was worried there will be more work involved to have it working! 🎉

src/shared/ui/components/select/select.component.tsx
```diff
@@ -47,7 +47,7 @@ export const Select = <T,>({
           <DropdownMenu.Content sideOffset={2} asChild>
             <SelectOptionContainer
               as="div"
-              className="w-[var(--radix-popper-anchor-width)] text-black"
+              className="max-h-64 w-[var(--radix-popper-anchor-width)] overflow-y-auto text-black"
```
nemmtor, line 50: Also for future cases it might be worth to add extra prop to this component which would allow consumers to change the max-height of options, I can imagine we can face a scenario where options should be taller/smaller etc.

You would need to add new prop to this component `optionsContainerClassName?: string` and merge these classes via our utility:
`className={classes('max-h-64 w-[var(--radix-popper-anchor-width)] overflow-y-auto text-black', optionsContainerClassName)}`

But we can also add it if we face this situation, that's minor and I'll leave it up to you if you want to add it or not.

tailwind.css
SolutionsEngineer: thanks, had no idea it's possible 🙌

src/shared/ui/components/select/select.component.tsx
SolutionsEngineer: Good idea, optionsContainerClassName prop added

nemmtor (approved):

lennardevertz (approved):

PR #58, opened 2024-08-27
feat: migrate to viem

nemmtor:


src/application/polymarket/types.ts
```diff
@@ -80,31 +76,83 @@ export interface PolymarketUser {
...
+  orderType: 'FOK';
+}
+
+export interface OrderData {
+  maker: string;
+  taker: string;
+  tokenId: string;
+  makerAmount: string;
+  takerAmount: string;
+  side: 0;
```
SolutionsEngineer, line 131: Are the values like 'BUY', 'FOK', '0', and '2' static, non-changing values?

src/application/polymarket/utils.ts
```diff
@@ -35,3 +61,275 @@ export const isEventUrl = (url: string) => {
...
+        type: 'uint256',
+      },
+    ],
+  } as const;
+
+  const message = {
+    address: address,
+    timestamp,
+    nonce: BigInt(0),
+    message: 'This message attests that I control the given wallet',
```
SolutionsEngineer, line 103: Just to make sure, was that message supposed to be a value or a comment?

src/application/polymarket/widgets/market-widget.tsx
```diff
@@ -209,8 +269,9 @@ export const Market = ({
                               <InputBase.Label label="Amount" />
                               <Chip className="bg-[#2C3F4F]">
                                 Balance $
-                                {user.safeWalletDetails?.balance.toFixed(2) ??
-                                  0}
+                                {(user.safeWalletDetails?.balance ?? 0).toFixed(
+                                  2,
+                                ) ?? 0}
```
SolutionsEngineer, line 274: I think we can remove this fallback as we have one in the toFixed source

src/shared/web3/types.ts
```diff
@@ -1,21 +1,22 @@
...
-  nativeCurrency: {
-    name: string;
-    symbol: string;
-    decimals: number;
-  };
-  rpcUrls: string[];
-  blockExplorerUrls: string[];
-}
-
+// export interface Chain {
```
SolutionsEngineer, line 6: Do we want to keep that interface commented?

SolutionsEngineer (approved):
I've left a few minor comments, but overall, it looks great 👌

src/application/polymarket/types.ts
nemmtor: it always has to be like that for our use-case, in general polymarket accepts some different values but we only care about these when making an order

src/application/polymarket/utils.ts
nemmtor: it was supposed to be a value, it is the one visible in metamask when you are asked to sign

src/application/polymarket/widgets/market-widget.tsx
nemmtor: good catch!

src/shared/web3/types.ts
nemmtor: hell no 😅

src/application/polymarket/types.ts
SolutionsEngineer: okay

SolutionsEngineer (dismissed):

PR #59, opened 2024-09-01
Extension popup menu

SolutionsEngineer:
**Governance Widgets:**

**Entire extension toggle:**

**Connect Wallet (Log In / Log Out):**

**Tipping:**

**Polymarket integration:**

**Look up wallet address:**

**Block explorers:**

**cross tab sync:**

https://github.com/user-attachments/assets/559c730f-b188-4313-9890-3753f59a3c12


package.json
```diff
@@ -41,6 +41,7 @@
     "ethers": "^5.6.9",
     "idriss-crypto": "^1.9.1",
     "lit": "^3.1.2",
+    "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
```
nemmtor, line 45: I am wondering if we need this granularity in lodash modules, maybe we can get whole lodash package and just import not via barell file so we don't get unncessary bundle.

`import debounce from 'lodash/debounce';`

src/application/extension-popup-menu/app.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { CustomizationSettings } from './settings-view/customization-settings';
+import { MainSettingsMenu } from './settings-view/main-settings-menu';
+
+export const EXTENSION_POPUP_MENU_ROUTES = {
+  HOME: '/',
+  SETTINGS: {
+    MAIN_VIEW: '/settings',
+    CUSTOMIZATION: '/settings/customization',
+  },
+};
```
nemmtor, line 18: I can imagine situation where we want to open given route by default from outside of popup, in this case we would need to also consume routes, we already discussed it on discord that it might be worth to open wallet connect modal inside popup, maybe we could move that const into `shared/extension` ?

src/application/extension-popup-menu/app.tsx
```diff
@@ -0,0 +1,55 @@
+import { Route, Routes } from 'react-router-dom';
+
+import { useExtensionSettings } from 'shared/extension';
+import { Closable } from 'shared/ui';
+
+import { Footer } from './footer';
+import { PopupHomeView } from './popup-home-view';
+import { TopBar } from './top-bar';
+import { CustomizationSettings } from './settings-view/customization-settings';
+import { MainSettingsMenu } from './settings-view/main-settings-menu';
```
nemmtor, line 10: missing barell file `import { CustomizationSettings, MainSettingsMenu } from './settings-view';`

src/application/extension-popup-menu/app.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const EXTENSION_POPUP_MENU_ROUTES = {
```
nemmtor, line 12: feels like `MENU` is unnecessary UI detail, `EXTENSION_POPUP_ROUTE` feels good enough, also note that I removed (s) to make it as a singular, it could be ROUTE(S) if it was an array but here you will be referring to a single route once consuming it: `EXTENSION_POPUP_ROUTE.HOME`

src/application/extension-popup-menu/app.tsx
```diff
@@ -0,0 +1,55 @@
...
+  HOME: '/',
+  SETTINGS: {
+    MAIN_VIEW: '/settings',
+    CUSTOMIZATION: '/settings/customization',
+  },
```
nemmtor, line 17: I am not sure about this nesting, usually nesting ends up in a more complex data structure than it needs to be, I suggest to do:
```ts
HOME: '/',
SETTINGS_HOME: '/settings',
SETTINGS_CUSTOMIZATION: '/settings/customization'
```

src/application/extension-popup-menu/app.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const App = () => {
+  const { isPopupMenuVisible, hidePopupMenu } = useExtensionSettings();
```
nemmtor, line 21: imho extension settings and popup should be decoupled, the fact that popup holds settings inside of it is just a detail that should not be important for consumers of popup, so I would recommend to do:
```ts
const { hide, open, isVisible } = useExtensionPopup();
```

it also looks nice when you're not destructuring:
```ts
const extensionPopup = useExtensionPopup();
extensionPopup.open(EXTENSION_POPUP_ROUTE.SETTINGS_CUSTOMIZATION);
```

src/application/extension-popup-menu/extension-popup-menu.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+      defaultValue: false,
+    },
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof WrappedComponent>;
+
+export const Primary: Story = {};
```
nemmtor, line 47: very nice 👏🏻

src/application/extension-popup-menu/footer.tsx
```diff
@@ -0,0 +1,46 @@
...
+      href: 'https://discord.gg/RJhJKamjw5',
+      iconName: 'DiscordLogoIcon',
+      text: 'Discord',
+    },
+    {
+      href: 'https://twitter.com/IDriss_xyz',
+      iconName: 'TwitterLogoIcon',
+      text: 'Twitter',
+    },
+  ];
```
nemmtor, line 26: this thing is static so could be moved outside of component (ideally constants file), interface also won't be needed since you're not expecting it as a prop but just consuming it directly in JSX

src/application/extension-popup-menu/footer.tsx
```diff
@@ -0,0 +1,46 @@
+import { Icon, IconName } from 'shared/ui';
+
+interface FooterLink {
+  iconName: IconName;
+  href: string;
+  text: string;
+}
+
+export const Footer = () => {
```
nemmtor, line 9: I suggest to create internal folder `components` and hold footer there to make it consistent with other places

src/application/extension-popup-menu/app.tsx
```diff
@@ -0,0 +1,55 @@
...
+          />
+          <Route
+            path={EXTENSION_POPUP_MENU_ROUTES.SETTINGS.MAIN_VIEW}
+            element={<MainSettingsMenu />}
+          />
+          <Route
+            path={EXTENSION_POPUP_MENU_ROUTES.SETTINGS.CUSTOMIZATION}
+            element={<CustomizationSettings />}
+          />
+        </Routes>
```
nemmtor, line 49: there is new recommended way of defining routes in react:
https://reactrouter.com/en/main/routers/create-browser-router

you could move this router definition outside to `router.tsx`:
```tsx
export const extensionPopupRouter = createBrowserRouter(...);
```

and inside here only do:
```ts
<RouterProvider router={extensionPopupRouter} />
```

but that's minor thing actually

nemmtor: one thing that I am worried about is that we won't be able to redirect popup to another view from another module like polymarket because you will be outside of react router context

Ideally I would want to do this inside polymarket:
```tsx
const logIn = () => {
  popup.open(EXTENSION_POPUP_ROUTE.LOGIN)
}

<Button onClick={login}>Connect wallet</Button>
```

Let's maybe discuss it internally in discord and leave for later

src/application/look-up-wallet-address/commands/index.ts
```diff
@@ -0,0 +1,10 @@
+import { GetResolvedAddressCommand } from './get-addresses.command';
+import { GetTwitterIdsCommand } from './get-twitter-ids.command';
+
+export const COMMAND_MAP = {
+  [GetTwitterIdsCommand.name]: GetTwitterIdsCommand,
+  [GetResolvedAddressCommand.name]: GetResolvedAddressCommand,
+};
```
nemmtor, line 7: we will have these in `shared/idriss`, it's currently part of different PR

src/application/extension-popup-menu/popup-home-view/app.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { SectionTile } from './components';
+
+export const App = () => {
+  const navigate = useNavigate();
+  return (
+    <div className="flex max-w-[570px] grow flex-col items-center bg-gray-100 px-1 pt-6">
+      <header className="mb-6 text-center">
+        <h1 className="mb-2 text-2xl font-bold text-black">
+          Hello, it&apos;s your web3 copilot speaking
```
nemmtor, line 22: cool trick to avoid `&apos;` and other things like that is to do:
```tsx
{`Hello, it's your web3 copilot speaking`}
```

src/application/extension-popup-menu/popup-home-view/app.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { TRADING_COPILOT_LOGO } from '../constants';
+import { EXTENSION_POPUP_MENU_ROUTES } from '../app';
+
+import { SectionTile } from './components';
+
+export const App = () => {
+  const navigate = useNavigate();
+  return (
+    <div className="flex max-w-[570px] grow flex-col items-center bg-gray-100 px-1 pt-6">
```
nemmtor, line 19: this 570px is already applied in parent, probably you don't need it here

src/application/extension-popup-menu/app.tsx
nemmtor: also I feel like `menu` can be removed from the module name `src/application/extension-popup` is good enough since there is only 1 extension popup

package.json
SolutionsEngineer: sure, makes sense 👌

src/application/extension-popup-menu/app.tsx
SolutionsEngineer: Great idea, looks much better 👌

src/application/extension-popup-menu/footer.tsx
SolutionsEngineer: I've moved the `footerLinks` to constants.ts along with the interface. I think we can keep it this way to make it easier to edit the `iconName` props in the future

src/application/extension-popup-menu/app.tsx
SolutionsEngineer: As discussed on Discord, I've replaced the HashRouter with the MemoryRouter and moved it higher in our Application hierarchy

src/application/extension-popup-menu/popup-home-view/app.tsx
SolutionsEngineer: thanks, replaced 👌

src/application/extension-popup-menu/popup-home-view/app.tsx
SolutionsEngineer: nice catch!

src/application/extension-popup/components/top-bar.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { IDRISS_ICON_WITH_TEXT } from 'shared/idriss';
+import { IconButton, Toggle } from 'shared/ui';
+import { EXTENSION_POPUP_ROUTE, useExtensionSettings } from 'shared/extension';
+
+export const TopBar = () => {
+  const { extensionSettings, changeExtensionSetting } = useExtensionSettings();
+  const navigate = useNavigate();
+  const { pathname } = useLocation();
+  const isHomeView = pathname === EXTENSION_POPUP_ROUTE.HOME;
```
nemmtor, line 11: I am a big fan of hiding implementation details, this would be a cool improvement to not couple whole repository with `react-router` interface.

To achieve it you would need to add it inside `usePopup` hook:
```ts
popup.navigate()
popup.currentRoute
```

and not use useNavigate and useLocation anywhere else in code, by doing so we will apply dependency inversion so we have a single interface owned by us instead of being ruled by a library.

On the other hand I know that we will probably never move away from this react-router, but this change shoulnd't be a lot of work, let me know WDYT, I am also fine with keeping react-router coupling

src/application/extension-popup/popup-home-view/components/section-tile.tsx
```diff
@@ -0,0 +1,71 @@
...
+              fillRule="evenodd"
+              d="M4.25 5.5a.75.75 0 0 0-.75.75v8.5c0 .414.336.75.75.75h8.5a.75.75 0 0 0 .75-.75v-4a.75.75 0 0 1 1.5 0v4A2.25 2.25 0 0 1 12.75 17h-8.5A2.25 2.25 0 0 1 2 14.75v-8.5A2.25 2.25 0 0 1 4.25 4h5a.75.75 0 0 1 0 1.5h-5Z"
+              clipRule="evenodd"
+            />
+            <path
+              fillRule="evenodd"
+              d="M6.194 12.753a.75.75 0 0 0 1.06.053L16.5 4.44v2.81a.75.75 0 0 0 1.5 0v-4.5a.75.75 0 0 0-.75-.75h-4.5a.75.75 0 0 0 0 1.5h2.553l-9.056 8.194a.75.75 0 0 0-.053 1.06Z"
+              clipRule="evenodd"
+            />
+          </svg>
```
nemmtor, line 66: maybe we can use external link icon from radix?

src/application/extension-popup/settings-view/customization-settings/app.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const { changeExtensionSetting, extensionSettings } = useExtensionSettings();
+  const navigate = useNavigate();
+  const isExtensionEnabled = extensionSettings['entire-extension-enabled'];
+
+  const settingListItemGroups: SettingListItemsGroup<ExtensionSettingsStorageKey>[] =
+    [
+      { label: 'Address Book', settingListItems: addressBookSettings },
+      { label: 'Governance', settingListItems: governanceSettings },
+      { label: 'Integrations', settingListItems: integrationsSettings },
+    ];
```
nemmtor, line 31: can be move to constants since it is static

src/shared/ui/components/checkbox/checkbox.types.ts
```diff
@@ -1,4 +1,19 @@
...
+} & (
+  | {
+      value: boolean;
+      type?: 'binary';
+    }
+  | {
+      type: 'extended';
+      value?: ExtendedCheckboxOption;
+    }
+);
```
nemmtor, line 17: can you explain why do we need that?

src/shared/ui/components/checkbox/checkbox.types.ts
```diff
@@ -1,4 +1,19 @@
-export interface CheckboxProperties {
-  value: boolean;
+export type CheckboxProperties = {
+  label?: string;
+  disabled?: boolean;
+  className?: string;
+  disabledTooltipText?: string;
+  additionalClassNameWhenChecked?: string;
```
nemmtor, line 6: this should be doable via className with `checked:` prefix and be injected via normal `className` property

src/application/extension-popup/app.tsx
```diff
@@ -0,0 +1,45 @@
+import { Route, Routes } from 'react-router';
+
+import { EXTENSION_POPUP_ROUTE, useExtensionPopup } from 'shared/extension';
+import { Closable } from 'shared/ui';
+
+import { Footer, TopBar } from './components';
+import { PopupHomeView } from './popup-home-view';
+import { CustomizationSettings, MainSettingsMenu } from './settings-view';
+
```
nemmtor, line 9: let's apply group these under `./views` and you can remove `-view` suffix, also inside each of them you can rename them to `CustomizationSettingsView`, `MainSettingsView` and `HomeView` instead of `App` + renaming in export

src/application/look-up-wallet-address/look-up-wallet-address.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import type { Meta, StoryObj } from '@storybook/react';
+
+import { LookUpWalletAddress } from '.';
```
nemmtor, line 3:
```suggestion
import { LookUpWalletAddress } from './app';
```

src/application/look-up-wallet-address/app.tsx
```diff
@@ -0,0 +1,121 @@
...
+        });
+      }
+    };
+
+    window.addEventListener('keydown', handleKeyDown);
+
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown);
+    };
+  }, []);
```
nemmtor, line 67: maybe you can experiment and see if we can use one of these:
https://github.com/streamich/react-use/blob/master/docs/useKey.md
https://github.com/streamich/react-use/blob/master/docs/useKeyPress.md
https://github.com/streamich/react-use/blob/master/docs/useKeyboardJs.md
https://github.com/streamich/react-use/blob/master/docs/useKeyPressEvent.md

src/application/look-up-wallet-address/app.tsx
```diff
@@ -0,0 +1,121 @@
...
+      setUsername(inputValue);
+    },
+    500,
+    [inputValue],
+  );
+
+  const handleInput = (event: FormEvent<HTMLInputElement>) => {
+    const eventTarget = event.target as HTMLInputElement;
+    setInputValue(eventTarget.value);
+  };
```
nemmtor, line 51: we can use react-hook-form + `watch` to get reactive value of input, I am also wondering if this debounce could not be replaced with <form> submit which submits on enter, again it's internal tool so we maybe we don't need that debounce?
CC @lennardevertz @geoist1

src/application/look-up-wallet-address/app.tsx
```diff
@@ -0,0 +1,121 @@
...
+export const App = () => {
+  const [username, setUsername] = useState<string>('');
+  const [inputValue, setInputValue] = useState<string>('');
+  const [visible, setVisible] = useState(false);
+
+  const { extensionSettings } = useExtensionSettings();
+
+  const isEnabled =
+    extensionSettings['entire-extension-enabled'] &&
+    extensionSettings['block-explorers-enabled'];
```
nemmtor, line 20: maybe `wallet-lookup-enabled` ?

src/application/look-up-wallet-address/app.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <label
+            htmlFor="first_name"
+            className="mb-1 block text-xs text-gray-700"
+          >
+            Look up your wallet address
+          </label>
+          <input
+            autoFocus
+            id="first_name"
+            className="mb-1.5 box-border block h-[38px] w-full rounded border border-[#cccccc] bg-white px-3 py-2 align-middle font-sans text-sm leading-[1.428571429] text-[#333333] outline-none"
```
nemmtor, line 91: this className feels bad :sweat:
`leading-[1.428571429]`

maybe we can reuse some of existing tailwind classnames or at least round it?

src/application/look-up-wallet-address/app.tsx
```diff
@@ -0,0 +1,121 @@
...
+          />
+          {inputValue.length > 0 && (
+            <IconButton
+              iconProps={{ name: 'Cross1Icon' }}
+              className="absolute right-2 top-1/2 text-black transition-transform hover:scale-90"
+              onClick={() => {
+                return setInputValue('');
+              }}
+            />
+          )}
```
nemmtor, line 106: you should add right padding to the input itself otherwise text can go under `X` button if it get's too long

src/application/look-up-wallet-address/app.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { useCommandQuery } from 'shared/messaging';
+import { Closable, IconButton } from 'shared/ui';
+import { useExtensionSettings } from 'shared/extension';
+
+import { GetResolvedAddressCommand, GetTwitterIdsCommand } from './commands';
+import { AddressList } from './address-list';
+
+export const App = () => {
+  const [username, setUsername] = useState<string>('');
+  const [inputValue, setInputValue] = useState<string>('');
```
nemmtor, line 13: you don't need to pass `<string>` generic, typescript understands type from default value that you've provided

src/shared/ui/components/tooltip/tooltip.component.tsx
```diff
@@ -0,0 +1,38 @@
+import { useEffect, useState } from 'react';
+
+import { TooltipProperties } from './tooltip.types';
+
+export const Tooltip = ({ children, tooltipMessage }: TooltipProperties) => {
```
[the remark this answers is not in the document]
nemmtor: for reference - it is used only to show here (Copied address!)
image

src/shared/ui/components/toggle/toggle.component.tsx
```diff
@@ -0,0 +1,29 @@
+import { classes } from 'shared/ui/utils';
+
+import { ToggleProperties } from './toggle.types';
+
+export const Toggle = ({ checked, onCheckedChange }: ToggleProperties) => {
```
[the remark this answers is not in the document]
nemmtor: btw. you also have this issue after fresh installation and triggering toggle for 1st time:
image

src/shared/ui/components/tooltip/tooltip.component.tsx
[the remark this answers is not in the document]
lennardevertz: totally fine, this copy does not really need a reaction at all imo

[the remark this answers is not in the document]
geoist1: @nemmtor ignore the comment above, just talked to @lennardevertz about this

although this feature won't be marketed, it will be more than merely an internal tool - we are going to communicate about it at least in the docs and to interested users

therefore, the UI & UX still needs to be good, but without investing too much time into it. 

1. let's either leave the tooltip as it is (totally fine), or if you insist @nemmtor let's display static text `Address copied ` in the list element + bg-green-100 applied to the background of the element for the same duration as the tooltip. We can also add closing of the dropdown after, as you suggested some time ago @nemmtor.
2. this made us realize that the copying functionality on block explorers (e.g. on Polygonscan.com) should not be there at all - the primary reason for it to exist there is to just fill the input field (as opposed to lookup widget where copying matters)

also, this comment made me do some quick visual improvements to the address lookup widget (in my PR): 6b95c8327759794ed63b7d6e0d00b4a343448e91 CC @SolutionsEngineer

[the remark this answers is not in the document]
nemmtor: yeah so let's just display some static text instead of implementing Tooltip because proper Tooltip would take too much time

src/application/look-up-wallet-address/app.tsx
nemmtor: my suggestion was to not call it before user presses enter/button

src/application/extension-popup/components/top-bar.tsx
SolutionsEngineer: sounds good, we won't have to add the `useNavigate`/`useLocation` imports everytime

SolutionsEngineer: I've also moved the MemoryRouter to the PopupContetxtProvider, let me know what do you think about that approach

nemmtor: sounds good!

src/application/extension-popup/popup-home-view/components/section-tile.tsx
SolutionsEngineer: good point

src/shared/ui/components/tooltip/tooltip.component.tsx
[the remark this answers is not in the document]
SolutionsEngineer: I've removed the tooltip and added static text replacing other content as suggested.
@geoist1 do we want to address the 2nd point in this PR?

src/shared/ui/components/checkbox/checkbox.types.ts
SolutionsEngineer: We are handling the intermediate state of the checkbox, when only a few of the section options are checked:
image

SolutionsEngineer: but right now I think we can just go for true | false | null and get rid of all of that additional code

SolutionsEngineer: I'll take a look on that

nemmtor: I think we should go with `true | false` and this "parent checkbox" could be a special case, maybe normal checkbox could take a prop `indicator: ReactNode` and you can decide to override default indicator (empty or tick icon) with whatever you want.

Something like:
```tsx
const showSomeItemCheckedIndicator = someItemInGroupChecked && !wholeGroupChecked;

// parent checkbox
<Checkbox
  value={wholeGroupChecked}
  onChange={toggleWholeGroupChange}
  indicator={showSomeItemCheckedIndicator ? <DashIcon /> : undefined}
/> 
```

src/shared/extension/utils.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  ExtensionSettingsStorageKey,
+  extensionSettingsStorageKeys,
+} from './extension-settings-manager';
+
+const initiallyDisabledExtensionSettingsStorageKeys: ExtensionSettingsStorageKey[] =
+  [] as const;
```
nemmtor, line 7: maybe we can just hardcode `true` few lines below in `createInitialExtensionSettingsStorageKeys`?

src/application/look-up-wallet-address/address-list.tsx
```diff
@@ -0,0 +1,103 @@
...
+        setCopiedAddressValue(undefined);
+        onAddressCopied();
+      }, 750);
+    }
+    return () => {
+      if (timeout) {
+        clearTimeout(timeout);
+      }
+    };
+  }, [copiedAddressValue, onAddressCopied]);
```
nemmtor, line 36: I asked GPT and seems like it can be done without useEffect:
```tsx
const handleAddressClick = useCallback((address: string) => {
  void navigator.clipboard.writeText(address);

  const timeout = setTimeout(() => {
    onAddressCopied();
  }, 750);

  // Clean up the timeout if this function is called again
  return () => {
    clearTimeout(timeout);
  };
}, [onAddressCopied]);
```

The only thing left which you will have `copiedAddress` state for is that green colour after copying, we already decided that UX/UI is not very important here so maybe we can get rid of it and remove state completely

SolutionsEngineer: I've decided to use the useEffect to allow copying other addresses in case of a misclick.
https://github.com/user-attachments/assets/b972345e-5949-465d-8bf9-27bf4b483bed

SolutionsEngineer: Do you think it can cause any problems?

src/shared/extension/utils.ts
SolutionsEngineer: sure, I just wanted to make it easier for the future, as I've heard that there can be some options that we would like to have initially disabled.

src/shared/ui/components/toggle/toggle.component.tsx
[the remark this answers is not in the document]
SolutionsEngineer: Thanks, it looks like using the Radix toggle fixed the problem

src/shared/ui/components/checkbox/checkbox.types.ts
SolutionsEngineer: thanks, I've refactored the checkbox component as we agreed on Discord

src/application/look-up-wallet-address/app.tsx
SolutionsEngineer: thanks, the useKey hook works great here

src/application/look-up-wallet-address/app.tsx
SolutionsEngineer: fixed, now it will can the API after the "Search" button click.
image
image

src/application/look-up-wallet-address/app.tsx
SolutionsEngineer: yeah... I've wanted to transform CSS to tailwind quickly using some 'smart' tool and forgot to remove that mess 😶‍🌫️

src/application/look-up-wallet-address/app.tsx
SolutionsEngineer: fixed
https://github.com/user-attachments/assets/1529c3ff-8d58-44c0-8f74-3c12b273a828

src/shared/extension/utils.ts
SolutionsEngineer: let me know if we want to keep that for the future, or just remove it for now

src/shared/ui/components/tooltip/tooltip.component.tsx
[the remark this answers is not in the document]
nemmtor: @geoist1

src/application/look-up-wallet-address/app.tsx
nemmtor: nice!

src/shared/extension/utils.ts
nemmtor: I am with with both 👍🏻 you decide

src/shared/extension/extension-settings-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+      });
+    });
+  },
+
+  getAllSettings(): Promise<Record<ExtensionSettingsStorageKey, boolean>> {
+    return new Promise((resolve) => {
+      void chrome.storage.local.get().then((result) => {
+        return resolve(result as Record<ExtensionSettingsStorageKey, boolean>);
+      });
+    });
```
nemmtor, line 78: one thing which worries me is that this is now pointing to everything which is stored inside localStorage, settings should be hidden under some field name like `settings` so:
```ts
chrome.storage.local.get(SETTINGS_STORAGE_KEY);
```

also I believe that this type is a bit lieing, what if this is a fresh user and he had never saved anything in chrome storage? Probably we won't get whole dictionary of settings, what we should probably do is something like this:
```tsx
     void chrome.storage.local.get().then((result) => {
        const settingsFromStorage = result as Partial<Record<ExtensionSettingsStorageKey, boolean>>;
        const settings = { ...DEFAULT_SETTINGS, ...settingsFromStorage }; // not sure if type here will be without Partial, if not let's type cast
        return resolve(settings);
      });
```

src/shared/extension/extension-settings-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+
+interface StoredWallet {
+  account: Hex;
+  providerRdns: string;
+}
+
+export const ExtensionSettingsManager = {
+  setSettings(settings: Partial<ExtensionSettings>) {
+    chrome.storage.local.set(settings).catch(console.error);
+  },
```
nemmtor, line 50: as mentioned in below comment let's hide it behind some storage key:
 ```ts
set({ SETTINGS_STORAGE_KEY:settings })
 ```

src/shared/extension/extension-settings-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  isEnabled(storageKey: ExtensionSettingsStorageKey): Promise<boolean> {
+    return new Promise((resolve) => {
+      chrome.storage.local.get([storageKey], (result) => {
+        const isEnabled = Boolean(result[storageKey] ?? true);
+        resolve(isEnabled);
+      });
+    });
+  },
```
nemmtor, line 88: my IDE is telling me that this1 is not being used but I am not sure, can you double check?

src/shared/extension/extension-settings-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+      void chrome.storage.local.get('idriss-wallet').then((storedWalletRaw) => {
+        const storedWallet = storedWalletRaw['idriss-wallet']
+          ? (JSON.parse(storedWalletRaw['idriss-wallet']) as StoredWallet)
+          : undefined;
+        return resolve(storedWallet);
+      });
+    });
+  },
+
+  getAllSettings(): Promise<Record<ExtensionSettingsStorageKey, boolean>> {
```
nemmtor, line 73: you already have type for that `ExtensionSettings`

src/shared/extension/extension-settings-manager.ts
```diff
@@ -0,0 +1,89 @@
...
+    return new Promise((resolve) => {
+      void chrome.storage.local.get().then((result) => {
+        return resolve(result as Record<ExtensionSettingsStorageKey, boolean>);
```
nemmtor, line 76: you already have type for that `ExtensionSettings`

src/infrastructure/constants.ts
```diff
@@ -0,0 +1,3 @@
+// eslint-disable-next-line boundaries/no-unknown-files
+export const EXTENSION_BUTTON_CLICKED = 'EXTENSION_BUTTON_CLICKED';
+export const ACTIVE_TAB_CHANGED = 'ACTIVE_TAB_CHANGED';
```
nemmtor, line 3: can you add a TODO to fix that

src/infrastructure/content-script/content-script.ts
```diff
@@ -83,28 +73,70 @@ export class ContentScript {
...
+    });
+
+    onWindowMessage<{ account: Hex; providerRdns: string }>(
+      'SAVE_WALLET',
+      (v) => {
+        void ExtensionSettingsManager.saveWallet(v);
+      },
+    );
+  }
 }
```
nemmtor, line 142: can you add TODO to move these message names to constants in shared/web3

src/final/hooks/use-prefetch-proposals.ts
```diff
@@ -60,20 +73,27 @@ export const usePrefetchProposals = ({ widgetData }: Properties) => {
...
 
-    if (hasSnapshotProposal) {
+    if (hasSnapshotProposal && snapshotEnabled) {
       sources.push('snapshot');
     }
 
-    if (hasTallyProposal) {
+    if (hasTallyProposal && tallyEnabled) {
       sources.push('tally');
     }
```
nemmtor, line 86: hm I think we can remove that 2nd condition since queries are disabled if given feature is disabled :think: but not sure what if we have some proposals and after that we change value in settings, is it the reason you've added it?

src/application/look-up-wallet-address/schema.ts
```diff
@@ -0,0 +1,5 @@
+import { z } from 'zod';
+
+export const twitterIdSchema = z.object({
+  twitterIDs: z.record(z.string(), z.string()),
```
nemmtor, line 4: I am not sure if this will always have value of a string, can you double check by trying to call this api with some non-existing username?
I remember one of our api's return empty string and other returns null or smth like that.

src/application/look-up-wallet-address/commands/get-twitter-ids.command.ts
```diff
@@ -0,0 +1,55 @@
...
+        throw new HandlerError(
+          'Something went wrong when fetching extension twitter Ids.',
+        );
+      }
+
+      const json = await response.json();
+
+      const validResponse = twitterIdSchema.parse(json);
+
+      return new OkResult(validResponse);
```
nemmtor, line 45: this1 will also consider single user so you might want to do `return new OkResult(validResponse[this.payload.username])` and response type will be `string`

src/application/look-up-wallet-address/look-up-wallet-address.tsx
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  useKey(
+    (event) => {
+      return event.ctrlKey && event.key === 'i';
+    },
+    () => {
+      return setIsVisible((previous) => {
+        return !previous;
+      });
```
nemmtor, line 62: maybe we can move to toggle function wrapped with useCallback and do:
```ts
useKey(() => { return event.ctrlKey && event.key === 'i' }, toggleVisibility)
```

even better would be:
```tsx
// utils.ts
const isShortcut = (event: KeyboardEvent) => {
  return event.ctrlKey && event.key === 'i';
}

// component
useKey(isShortcut, toggleVisibility);
```

src/application/look-up-wallet-address/look-up-wallet-address.tsx
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  const form = useForm();
+
+  const userIdentifier = form.watch('user_identifier');
+
+  const onSubmit = () => {
+    const values = form.getValues();
+    setUsername(values.user_identifier);
+  };
```
nemmtor, line 47: with react-hook-form you don't need to manually call getValues, your submit handler will receive data from form, you should also add correct type:
```tsx
type LookupFormValues = {
  identifier: string;
}

const EMPTY_LOOKUP_FORM_VALUES: LookupFormValues = {
  identifier: '',
};

const form = useForm<LookupFormValues>({ defaultValues: EMPTY_LOOKUP_FORM_VALUES });

const onSubmit: SubmitHandler<LookupFormValues> = (data) => {
  setUsername(data);
}
```

src/application/look-up-wallet-address/look-up-wallet-address.tsx
```diff
@@ -0,0 +1,127 @@
...
+        ? Object.values(twitterIdsQuery.data?.twitterIDs)[0]
+        : undefined,
+    }),
+    staleTime: Number.POSITIVE_INFINITY,
+    enabled: username.length >= 3,
+  });
+
+  const form = useForm();
+
+  const userIdentifier = form.watch('user_identifier');
```
nemmtor, line 42: let's use camelCase

src/application/look-up-wallet-address/look-up-wallet-address.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const isEnabled =
+    extensionSettings['entire-extension-enabled'] &&
+    extensionSettings['wallet-lookup-enabled'];
+
+  const twitterIdsQuery = useCommandQuery({
+    command: new GetTwitterIdsCommand({
+      username,
+    }),
+    staleTime: Number.POSITIVE_INFINITY,
+  });
```
nemmtor, line 27: should this1 be also disabled if username is less than X?

src/application/extension-popup/app.tsx
```diff
@@ -0,0 +1,43 @@
+import { Route, Routes } from 'react-router';
+
+import { EXTENSION_POPUP_ROUTE, useExtensionPopup } from 'shared/extension';
+import { Closable } from 'shared/ui';
+
+import { Footer, TopBar } from './components';
+import { HomeView } from './views/home/home';
+import { MainSettingsView } from './views/main-settings/main-settings';
+import { CustomizationSettingsView } from './views/customization-settings/customization-settings';
```
nemmtor, line 9: can we re-export these through barell files?
so:
```ts
import { HomeView, MainSettingsView, CustomizationSettingsView } from './views';
```

src/application/extension-popup/views/main-settings/main-settings-menu-item.tsx
```diff
@@ -0,0 +1,38 @@
...
+}: MainSettingsMenuListItem) => {
+  return (
+    <li
+      className="flex shrink-0 grow cursor-pointer items-center space-x-2 whitespace-nowrap py-3 text-base hover:text-green-500"
+      onClick={
+        onClick &&
+        (() => {
+          return onClick();
+        })
+      }
```
nemmtor, line 26: I believe you can do: `onClick={onClick}`

src/application/extension-popup/views/main-settings/main-settings-menu-item.tsx
```diff
@@ -0,0 +1,38 @@
+import { ReactNode } from 'react';
+
+import { Icon, IconButton, IconName } from 'shared/ui';
+
+interface MainSettingsMenuListItem {
+  label: ReactNode;
+  prefixIconName: IconName;
+  suffixIconName?: IconName;
```
nemmtor, line 8: could be easier if you do:
```ts
prefix?: ReactNode;
suffix?: ReactNode;

// ...
{prefix}
<span>{label}</span>
{suffix}
```
and you just inject icons from parent

src/application/extension-popup/views/home/components/section-tile.tsx
```diff
@@ -0,0 +1,65 @@
...
+      target={href ? '_blank' : undefined}
+      rel={href ? 'noopener noreferrer' : undefined}
+      className={classes(
+        'relative rounded-lg p-4 text-center shadow-md',
+        !disabled &&
+          'transition-transform duration-[600ms] hover:scale-105 hover:bg-gray-50',
+        disabled && 'cursor-not-allowed bg-gray-200 opacity-50',
+      )}
+    >
+      <div
```
nemmtor, line 37: I believe that SectionTile should be unaware of being wrapped with a link and in the places where you want it to be a link you can just wrap it with `<a ...>` by doing so we will avoid having weird anchors without links etc. and you won't need to have `disabled` state for anchor which in reality there's nothing like that in HTML.

src/application/extension-popup/views/home/home.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { SectionTile } from './components';
+
+export const HomeView = () => {
+  const extensionPopup = useExtensionPopup();
+
+  return (
+    <div className="flex max-w-[470px] grow flex-col items-center bg-gray-100 px-1 py-4">
+      <header className="mb-4 text-center">
+        <h1 className="mb-1 text-xl font-bold text-black">
+          GM, it&apos;s your web3 copilot speaking
```
nemmtor, line 21: same as in other place we can do `{`GM, it's ...`}`

src/application/extension-popup/views/customization-settings/customization-settings.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const customizationSettings = useMemo(() => {
+    const { 'entire-extension-enabled': _, ...customizationSettings } =
+      extensionSettings;
+
+    return customizationSettings;
+  }, [extensionSettings]);
+
+  const previousFormValues = useRef<ExtensionSettingsFormValues>(
+    customizationSettings,
+  );
```
nemmtor, line 39: can you please add descriptive comment why do we need this?
I will probably forgot why we need this workaround in a couple of days :P

src/application/look-up-wallet-address/look-up-wallet-address.tsx
```diff
@@ -0,0 +1,127 @@
...
+            </label>
+            <div className="flex flex-row gap-x-[10px]">
+              <div className="relative grow">
+                <input
+                  autoFocus
+                  id="user_identifier"
+                  className="mb-1.5 box-border block h-[38px] w-full rounded border border-[#cccccc] bg-white py-2 pl-3 pr-10 align-middle font-sans text-sm text-[#333333] outline-none"
+                  placeholder="hello@idriss.xyz  |  +1 650...  |  @IDriss_xyz"
+                  {...form.register('user_identifier')}
+                />
```
nemmtor, line 92: if you will try to find `register(` in repo you will only find this1, that's because in other places I use `Controller` component which is more flexible:
```tsx
<Controller
  render={({ field }) => {
   return <input {...field}> 
  }
 >
```

`field` under the hood includes things like `value`, `onChange`, etc.

src/infrastructure/constants.ts
nemmtor: actually you can move it to shared/extension constants

src/application/extension-popup/views/home/components/section-tile.tsx
nemmtor: also this new SectionTile does not need to accept `disabled` prop, you can just inject `className` prop with specific styles for trading co-pilot

src/shared/ui/components/tooltip/tooltip.component.tsx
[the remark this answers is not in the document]
lennardevertz: 2. fixed in [294ed1d](https://github.com/idriss-crypto/browser-extensions/pull/59/commits/294ed1d6a62fb7ec5184a9305cf93530cd2627d3)

src/application/extension-popup/views/customization-settings/customization-settings.tsx
SolutionsEngineer: good idea

src/application/look-up-wallet-address/schema.ts
SolutionsEngineer: in this case it just returns an empty object

```twitterIDs: {}```

File: src/final/hooks/use-prefetch-proposals.ts (hunk 38 above)
Comment (SolutionsEngineer) [reply src/final/hooks/use-prefetch-proposals.ts:86]: exactly, otherwise the widgets do not disappear when changing the settings after the proposals have already been fetched

File: src/shared/extension/context/popup.tsx [hunk 52]
```diff
@@ -0,0 +1,97 @@
...
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const toggleVisibility = useCallback(
+    debounce(() => {
+      setIsVisible((previous) => {
+        return !previous;
+      });
+    }, 50),
+    [],
+  );
```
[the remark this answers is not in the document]
SolutionsEngineer: I'm not sure if useDebounce from react-use will work here, as it only calls the passed function when the dependency array changes, and does not create a callable debounced function like debounce from lodash. However, I haven't used it much, so I could be wrong.

We can discuss whether we actually need the debounce here. I added it for two reasons:

1. A double-click makes it appear.
2. In React.StrictMode, scripts are injected twice, which prevents the modal from toggling correctly. I believe this could be fixed by adding removeEventListener in the useEffect cleanup

nemmtor (approved):

PR #61, opened 2024-09-02
feat: migrate idriss send to warpcast

nemmtor:


src/popup/extension-status-manager.ts
```diff
@@ -2,33 +2,32 @@
...
 
+const STORAGE_KEY = 'extension-status';
 export class ExtensionStatusManager {
-  private static STORAGE_KEY = 'extension-status';
-
   static enable() {
     chrome.storage.local
       .set({
-        [this.STORAGE_KEY]: true,
+        [STORAGE_KEY]: true,
```
nemmtor, line 10: @SolutionsEngineer I removed `this` in this class because at some point in runtime I saw that it was showing `undefined`, maybe due to the fact that it is static class.

src/shared/farcaster/api.ts
```diff
@@ -0,0 +1,9 @@
+export const getUserFid = (username: string) => {
+  return fetch(
+    `https://fnames.farcaster.xyz/transfers/current?name=${username}`,
+  );
+};
+
+export const getAccountByFid = (fid: number) => {
+  return fetch(`https://api.idriss.xyz/get-connected-addresses?fid=${fid}`);
+};
```
nemmtor, line 9: @SolutionsEngineer WDYT about making this kind of `api` files instead of calling fetch directly in command handlers?

Previously I thought command handlers will only be responsible for calling api's but in reality they often do some extra logic.

Imho we could have these descriptive "api methods" in `api.ts` file in each module that defines some requests and probably we could have schema validation inside so we have type-safety and good DX for return type of these api methods.

src/host/warpcast/scraper.ts
```diff
@@ -64,8 +66,144 @@ export class Scraper {
...
+    const usersFromSuggestedFollow = Scraper.getSuggestedFollowsSectionUsers();
+    const handleUser = Scraper.getHandleUser();
+
+    if (handleUser) {
+      return [...usersFromPosts, ...usersFromSuggestedFollow, handleUser];
+    }
+
+    return [...usersFromPosts, ...usersFromSuggestedFollow];
   }
 }
```
nemmtor, line 209: @SolutionsEngineer in this file I took a bit better (imo) approach for scraping - I made composition of well named sub-methods so it is easier to follow what is going on.

Imho it's much better than what we currently have on Twitter which is some crazy query selector that no one understands 😅

SolutionsEngineer: Yeah, it was really nice to review that, and I think it will pay off if something changes in the future. It will be much easier to identify which function stops working. Great approach.

src/shared/farcaster/api.ts
SolutionsEngineer: Agree, moving the schema validation to the API methods would also reduce some code in the command files, making them easier to read if additional logic is added. I'm only afraid that the API methods file might grow too large, but since an api.ts file would be created for each module, I don’t think it will get messy

SolutionsEngineer (approved):
Nice, I really like the improvements to the scraper and the JSDoc you've added 🙌. Now it can be read like a novel 👌👌😄

SolutionsEngineer (approved):

PR #65, opened 2024-09-18
scaling down default screen + other visual improvements

geoist1:
small visual improvements to the default screen


src/application/extension-popup/popup-home-view/app.tsx
```diff
@@ -16,37 +16,44 @@ import { SectionTile } from './components';
...
           ]}
           title="Governance Widget"
-          subtitle="Get reminders about active DAO votes"
+          subtitle="Get alerts about active DAO votes"
+          iconContainerClassName="mt-2"
         />
 
         <SectionTile
           href="https://x.com/IDriss_xyz/status/1780582103396266085"
+          target="_blank"
```
SolutionsEngineer, line 55: We can get rid of the target and rel here. The ones added inside for the <a /> tag should be enough

src/application/extension-popup/popup-home-view/components/section-tile.tsx
```diff
@@ -18,26 +20,28 @@ export const SectionTile = ({
...
+      <div className={classes('mb-2 flex justify-center space-x-3', iconContainerClassName)}>
         {icons.map((icon) => {
           return (
             <img
               key={icon.iconSrc}
               src={icon.iconSrc}
               alt={icon.iconAlt}
-              className="size-8 rounded-lg"
-            />
+              className={`${icon.className ? icon.className : 'w-8 h-8'} rounded-lg`}            />
```
SolutionsEngineer, line 44: Can we use classes function here? I believe this function solves some tailwind problems with conditional classNames.
```className={classes('h-8 w-8 rounded-lg', icon.className)}```

File: src/application/extension-popup/popup-home-view/app.tsx [hunk 3]
```diff
@@ -16,37 +16,44 @@ import { SectionTile } from './components';
...
 
         <SectionTile
           href="https://x.com/IDriss_xyz/status/1761049574691819798"
+          target="_blank"
+          rel="noopener noreferrer"
           icons={[
-            { iconSrc: AGORA_LOGO, iconAlt: 'Agora Logo' },
-            { iconSrc: SNAPSHOT_LOGO, iconAlt: 'Snapshot Logo' },
-            { iconSrc: TALLY_LOGO, iconAlt: 'Tally Logo' },
+            { iconSrc: AGORA_LOGO, iconAlt: 'Agora Logo', className: 'w-6 h-6' },
```
SolutionsEngineer, line 44: for the same width and height, we can use the size prop
```size-6```

src/application/extension-popup/popup-home-view/app.tsx
geoist1: ok, I added it as I think opening in new tab wasnt working without it

PR #68, opened 2024-09-23
fix: type safe service worker

nemmtor:


src/shared/observability/commands/send-to-sentry.ts
```diff
@@ -1,16 +1,20 @@
...
   fetchOptions?: RequestInit;
-}
+};
 
-interface Response {
+type Response = {
   status: number;
-  headers: Headers;
-}
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
SolutionsEngineer, line 12: do we need that here?

src/shared/web3/commands/get-across-chain-fee.ts
```diff
@@ -15,14 +15,15 @@ interface Payload {
   amount: string;
   message: string;
   recipient: string;
-}
+};
 
-interface Response {
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
SolutionsEngineer, line 20: I think we can get rid of that since the rule is off

SolutionsEngineer (approved):

PR #70, opened 2024-09-26
fix: sentry spam

nemmtor:


prettier.config.mjs
```diff
@@ -10,6 +10,5 @@ export default {
   tabWidth: 2,
   printWidth: 80,
   arrowParens: 'always',
-  editorconfig: true,
```
nemmtor, line 13: removed as this rule is actually unknown by prettier, it got revealed after running prettier in cli

src/shared/observability/scope.tsx
```diff
@@ -102,24 +91,9 @@ interface Properties {
...
-      scope.captureException(event.error);
-    };
-    self.addEventListener('error', captureException);
-
-    return () => {
-      self.removeEventListener('error', captureException);
-    };
-  }, [scope]);
+    return createObservabilityScope('injected-app');
+  }, []);
```
nemmtor, line 96: this was reporting all errors that were appearing on window object so I think it was captuing errors from outside of our extension

src/shared/ui/providers/query.tsx
```diff
@@ -1,48 +1,13 @@
...
-        onError: (error, query) => {
-          observabilityScope.setContext('query', {
-            queryHash: query.queryHash,
-          });
-          observabilityScope.setFingerprint([
-            query.queryHash.replaceAll(/\d/g, '0'),
-          ]);
-          observabilityScope.captureException(error);
-        },
-      }),
```
nemmtor, line 45: this was doubling the errors as api calls errors are already reported in command handlers, trade-off of removing this is that we won't catch errors from metamask but as discussed on Discord we will try to incrementally adopt on errors from metamask in order not to get blocked again

src/shared/observability/scope.tsx
```diff
@@ -27,14 +21,10 @@ const integrations = getDefaultIntegrations({}).filter((defaultIntegration) => {
...
 
-const createClient = (
-  executionEnvironment: SentryExecutionEnvironment,
-  extensionId: string,
-) => {
+const createClient = (executionEnvironment: SentryExecutionEnvironment) => {
   return new BrowserClient({
     dsn: process.env.SENTRY_DSN,
     enabled: process.env.SENTRY_ENVIRONMENT === 'production',
-    allowUrls: [`chrome-://${extensionId}`],
```
nemmtor, line 37: I removed it because:
1. it was not working - it needs InboundFilter Sentry integration but when dealing with chrome extensions some of the integrations are not working or even should be disabled (sentry docs)
2. errors caught in react-lifecycle are not coming from `chrome-...` but from normal url's like `twitter.com`

SolutionsEngineer (approved):

PR #87, opened 2024-10-10
feat: idriss send on farcaster

nemmtor:


SolutionsEngineer (dismissed):

src/final/hooks/use-user-widgets.ts
```diff
@@ -31,17 +49,27 @@ export const useUserWidgets = () => {
     enabled: gitcoinEnabled && isTwitter,
   });
 
-  if (isWarpcast) {
-    return {
-      // widgets: userWidgetDataAdapter.fromScrapedUsers({ users }), // TODO: uncomment to enable user widgets on Warpcast
-      widgets: [],
-    };
-  }
+  if (isWarpcast || isSupercast) {
```
SolutionsEngineer, line 52: what do you think about moving this condition to some variable and reusing it in the query and here?

src/host/warpcast/scraper.ts
```diff
@@ -147,62 +174,350 @@ export class Scraper {
...
+    return {
+      node,
+      top: rect.top + window.scrollY,
+      data: {
+        username,
+      },
+    };
+  }
+
+  // private static getUserFromHoverCardContent() {
```
SolutionsEngineer, line 465: is that something we want to keep?

SolutionsEngineer (approved):

PR #91, opened 2024-10-20
feat: badges for farcaster followers

nemmtor:


SolutionsEngineer (approved):

SolutionsEngineer (approved):
Awesome idea with the key stored in the element data tag 👌

PR #97, opened 2024-10-26
fix: use alternative regex to get polymarket slug

lennardevertz:


apps/extension/src/application/polymarket/commands/get-condition-id.ts
```diff
@@ -88,8 +88,13 @@ export class GetConditionIdCommand extends Command<Payload, string> {
...
     const regex =
       /<meta\s+name="twitter:image"\s+content="[^"]*mslug=([^"&]+)"/i;
+    const regex2 =
+      /<meta\s+property="og:image"\s+content="[^"]*mslug=([^"&]+)(?:&|&amp;)?/i;
 
-    const match = regex.exec(htmlString);
+    let match = regex.exec(htmlString);
+    if (!match || match.length === 0) {
+      match = regex2.exec(htmlString);
+    }
```
nemmtor, line 97: can be simplified, I also suggest to apply descriptive name:
```ts
const twitterImageRegex = ...;
const ogImageRegex = ...;

const match = twitterImageRegex.exec(htmlString) ?? ogImageRegex.exec(htmlString);
return match?.[1]
```

lennardevertz: good one, fixed!

nemmtor (approved):

PR #100, opened 2024-10-27
Feat/hero section

SolutionsEngineer:
https://github.com/user-attachments/assets/3101fc1b-36fa-4bdc-8515-61c682fa5c04

https://github.com/user-attachments/assets/c2baf328-05f4-4be4-9069-94ae70253dcc


apps/main-landing/src/components/hero-section/components/tokens-flow-image.tsx
```diff
@@ -0,0 +1,8 @@
+export const TokensFlowImage = () => {
+  return (
+      <img
+        src="/tokens-flow.png"
+        className="min-w-[500px] -mt-[22%] md:-mt-[18%] lg:-mt-[15%] w-full z-0"
+      />
+  );
+};
```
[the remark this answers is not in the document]
nemmtor: also we should use `priority` prop as this will be LCP

apps/main-landing/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "next-env.d.ts",
     "**/*.ts",
     "**/*.tsx",
-    ".next/types/**/*.ts"
+    ".next/types/**/*.ts",
```
nemmtor, line 20: comma at the end is invalid json

packages/tailwind-config/tailwind.config.mjs
```diff
@@ -95,12 +95,16 @@ export default {
...
+        nav: '50',
       },
       fontFamily: {
         sans: ['var(--font-aeonikpro)'],
       },
       spacing: {
         4.5: '18px',
+        5.5: '22px',
+        26: '104px',
+        50: '200px',
```
nemmtor, line 107: I believe these could be inlined as probably it's only a single use-case so no need to pre-define it in tw config

packages/ui/src/components/branding-assets/snapshot-full.tsx
```diff
@@ -0,0 +1,36 @@
+type Properties = {
+  className?: string;
+};
+
+export const SnapshotFullLogo = ({ className }: Properties) => {
+  return (
+    <svg
+      className={className}
+      width="590px"
+      height="126px"
```
nemmtor, line 10: other logos dont have px suffix

packages/ui/src/components/marquee/marquee.tsx
```diff
@@ -0,0 +1,57 @@
+import { Children, ReactNode } from 'react';
+
+import { classes } from '../../utils';
+
+interface MarqueeProperties {
```
nemmtor, line 5: let's use type and name it `Properties`

packages/ui/src/components/marquee/marquee.tsx
```diff
@@ -0,0 +1,57 @@
...
+  children: ReactNode;
```
nemmtor, line 6: maybe we can accept:
`items: ReactNode[]` so you don't have to use `Children.map` but normal map

packages/ui/src/components/marquee/marquee.tsx
```diff
@@ -0,0 +1,57 @@
...
+        className,
+      )}
+    >
+      {/* Duplicate for seamless looping */}
+      {[0, 1, 2].map((duplicate) => {
+        return (
+          <div
+            key={duplicate}
+            className={classes(
+              'inset-0 mx-0 flex flex-none animate-marquee gap-x-0 whitespace-nowrap lg:mx-5 lg:gap-x-10',
```
nemmtor, line 32: mx-0 is probably unnecessary

nemmtor: why do we need this flex-none?

packages/ui/src/components/marquee/marquee.tsx
```diff
@@ -0,0 +1,57 @@
...
+              pauseOnHover &&
+                'group-focus-within:paused-animation group-hover:paused-animation',
+            )}
+            style={{
+              animationDuration: `${speed}s`,
+              transform: `translateX(${duplicate * 100}%)`,
+            }}
+            role="list"
```
nemmtor, line 40: maybe instead of role="list" you can use `ul` tag and `li` inside loop, just make sure to use `list-none` classname

apps/main-landing/src/components/hero-section/components/background-lines.tsx
```diff
@@ -0,0 +1,14 @@
...
+    <Image
+      priority
+      width={1820}
+      height={3702.656}
+      src="/hero-background-lines.png"
+      className="absolute top-0 hidden opacity-40 lg:block"
+      alt=""
+    />
+  );
+};
```
nemmtor, line 14: I solved issue of type-check in CI so you can move back to locally imported images, by doing so you don't need to apply width and height props since Nextjs detects them automatically

apps/main-landing/src/components/hero-section/components/collaborations-showcase.tsx
```diff
@@ -0,0 +1,100 @@
...
+    link: 'https://x.com/IDriss_xyz/status/1750851248121930167',
+  },
+  {
+    logo: <TallyLogo className={logoClassName} />,
+    link: 'https://x.com/IDriss_xyz/status/1807798872930202007',
+  },
+  {
+    logo: (
+      <AgoraFullLogo
+        className={classes(logoClassName, 'h-[17.42px] lg:h-[31.7px]')}
```
nemmtor, line 54: does it work to have values after commas in px values? :D maybe it should be 18/32 or 17/32?

apps/main-landing/src/components/hero-section/components/collaborations-showcase.tsx
```diff
@@ -0,0 +1,100 @@
...
+  ParallelLogo,
+  PolymarketLogo,
+  SnapshotFullLogo,
+  TallyLogo,
+} from '@idriss-xyz/ui/branding-assets';
+import { Button } from '@idriss-xyz/ui/button';
+import { Marquee } from '@idriss-xyz/ui/marquee';
+import { classes } from '@idriss-xyz/ui/utils';
+
+type CollaborationsShowcaseProperties = {
```
nemmtor, line 15: let's name it simple `Properties` sinc we're not reusing it outside anyway

apps/main-landing/src/components/hero-section/components/collaborations-showcase.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {collaborationLinks.map((collaborationLink) => {
+          return (
+            <Button
+              key={collaborationLink.link}
+              intent="tertiary"
+              size="large"
+              asLink
+              target="_blank"
+              href={collaborationLink.link}
+            >
```
nemmtor, line 92: let's add `tabIndex={-1}` to make sure it's not focusable via keyboard

apps/main-landing/src/components/hero-section/components/background-lines.tsx
nemmtor: also given this you probably don't need separate component as it's redundant abstraction, doing this feels good enough, no need to abstract to separate component in my opinion:
```ts
    <Image
      priority
      src={backgroundLines}
      className="absolute top-0 hidden opacity-40 lg:block"
      alt=""
    />
```

apps/main-landing/src/components/hero-section/components/tokens-flow-image.tsx
```diff
@@ -0,0 +1,14 @@
...
+    <Image
+      priority
+      width={1985}
+      height={1240}
+      src="/hero-tokens-flow.png"
+      className="z-0 -mt-[22%] w-full min-w-[500px] md:-mt-[18%] lg:-mt-[15%]"
+      alt=""
+    />
+  );
+};
```
nemmtor, line 14: same as here
https://github.com/idriss-crypto/browser-extensions/pull/100#discussion_r1826986050

nemmtor: there's eslint warning:
```
Arbitrary value classname 'lg:-mt-[15%]' should not start with a dash (-)eslinttailwindcss/enforces-negative-arbitrary-values
```

to get rid of it you should use `lg:mt-[-15%]` so specify `-` sign inside `[]` not outside

apps/main-landing/src/components/hero-section/components/collaborations-showcase.tsx
SolutionsEngineer: Yeah, it's working. Wdyt about keeping px for simplicity?
image

nemmtor (approved):

PR #107, opened 2024-10-30
feat: tertiary button

nemmtor:


packages/ui/src/components/button/variants.ts
```diff
@@ -30,6 +36,18 @@ export const button = cva(
         true: ['pr-3'],
       },
     },
+    compoundVariants: [
```
SolutionsEngineer, line 39: shouldn't it be something like this?
```
compoundVariants: [
      {
        size: 'medium',
        withPrefixIcon: true,
        className: 'py-3.5',
      },
      {
        size: 'medium',
        withSuffixIcon: false,
        className: 'py-3',
      },
    ],

PR #130, opened 2024-11-10
feat: reuse modal from design system

nemmtor:
https://github.com/user-attachments/assets/d1bbbacb-402e-4a6a-9baf-09a82f3e29b1

### DIFF
--- a/packages/ui/src/components/modal/modal.tsx
+++ b/packages/ui/src/components/modal/modal.tsx
@@ -21,6 +21,7 @@ type Properties = {
   left?: number;
   right?: number;
   top?: number;
+  headerContainerClassName?: string;
   withoutPortal?: boolean;
 };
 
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/modal/modal.tsx
```diff
@@ -19,23 +30,66 @@ export const Modal = ({
...
+  const shouldBeCentered = useMemo(() => {
+    return [left, right, top].every((value) => {
+      return value === undefined;
+    });
+  }, [left, right, top]);
+
+  const Container = useCallback(
+    (properties: { children: ReactNode }) => {
+      if (withoutPortal) {
+        return <>{properties.children}</>;
```
SolutionsEngineer, line 54: I think we can remove React Fragment from here

SolutionsEngineer (approved):

PR #134, opened 2024-11-10
feat: tipping form rebrand

nemmtor:


apps/extension/src/shared/idriss/components/send/form.tsx
```diff
@@ -39,23 +51,66 @@ export const Form = ({
...
+                  field.onChange(Number(value));
+                }}
+                label="Amount ($)"
+                numeric
               />
             );
           }}
         />
+
+        <div className="mt-2 grid grid-cols-3 gap-4">
```
SolutionsEngineer, line 69: wdyt about using creating a constant like `predefinedAmounts = [1,2,5]` and using `predefinedAmounts.map((amount => ....`?

packages/ui/src/components/form/field.tsx
```diff
@@ -0,0 +1,75 @@
...
+            className={classes(
+              'flex space-x-1 text-label7 text-neutral-600 lg:text-label6',
+              inputProperties.error && 'text-red-500',
+            )}
+          >
+            {inputProperties.error && (
+              <Icon name="AlertCircle" size={12} className="p-0.5" />
+            )}
+            {helperText}
+            {!inputProperties.error && (
```
SolutionsEngineer, line 65: I think this one should be `HelpCircle`

apps/extension/src/shared/idriss/components/send/form.tsx
nemmtor: I added TODO to tackle this later

SolutionsEngineer (approved):

PR #147, opened 2024-11-11
- update waiting screen and metadata

lennardevertz:


apps/main-landing/src/app/layout.tsx
```diff
@@ -83,8 +83,8 @@ export default async function RootLayout({
     return (
       <html lang="en">
         <body className={`${aeonikPro.variable} font-sans antialiased`}>
-          <div className="flex h-screen items-center justify-center bg-mint-900 text-white">
-            <h1 className="text-heading1">Soon...</h1>
+          <div className="flex h-screen items-center justify-center bg-mint-400 text-[#fafff5]">
```
SolutionsEngineer, line 86: we can use text-mint-100 here

PR #182, opened 2024-12-05
feat: add checking if extension is outdated

MateuszStawski:


lennardevertz:
Closing as it does not solve the problem

PR #193, opened 2024-12-09
Fix: Sending walletConnect transactions fail on first attempt

daniel0ar:
## Overview
Fixes an error where transactions where failing after connecting a mobile wallet through WalletConnect and sending a transaction, without any feedback to the user as well.

# How it was fixed
- Estimating gas was failing due to using a publicClient on Ethereum mainnet instead of using the connected walletClient. The files `use-erc20-transaction.ts`, `use-native-transaction.ts`, `use-sender.ts` and finally `content.ts` were refactored to use only the walletClient from the connected wallet (and removed all publicClient code).
- Error handling was added for some general cases, tho more error handling should be added to show feedback to the user for specific flows like rejecting a transaction.

## Checklist
- [x] I have tested rejecting transactions.
- [x] I have tested on not enough funds transactions.
- [x] I have tested sending transactions.


apps/main-landing/src/app/creators/donate/content.tsx
```diff
@@ -206,11 +209,15 @@ export const Content = ({ className }: Properties) => {
...
-        recipientAddress: validAddress,
-      });
+
+      try {
+        await sender.send({
+          sendPayload,
+          recipientAddress: validAddress,
+        });
+      } catch (error) {
+        console.error('Uknown error sending transaction.', error);
```
lennardevertz, line 219: Typo

lennardevertz:
Looks good, just a small typo in the error message

apps/main-landing/src/app/creators/donate/content.tsx
daniel0ar: Solved

PR #199, opened 2024-12-15
add API

MaciejKutynia:


apps/copilot-api/entities/subscribtions.entity.ts
```diff
@@ -0,0 +1,21 @@
+import {Entity, JoinColumn, ManyToOne, PrimaryColumn} from "typeorm";
+import {SubscribersEntity} from "./subscribers.entity";
+import {AddressesEntity} from "./addreesses.entity";
+
+@Entity('subscriptions')
+export class SubscriptionsEntity {
+    @PrimaryColumn({type: 'text', name: 'subscriber_id'})
+    readonly subscriber_id!: string;
```
daniel0ar, line 8: No need for suscription_id?

MaciejKutynia: I copy this from existing table ant there are only 2 columns

CREATE TABLE IF NOT EXISTS subscriptions (
  subscriber_id TEXT REFERENCES subscribers(subscriber_id),
  address TEXT REFERENCES addresses(address),
  PRIMARY KEY (subscriber_id, address)
);

apps/copilot-api/knexfile.js
```diff
@@ -0,0 +1,20 @@
+// knexfile.js
```
daniel0ar, line 1: Not needed

apps/copilot-api/knexfile.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = {
+    development: {
+        client: "pg",
+        connection: process.env.DATABASE_URL,
+        migrations: {
+            directory: "./migrations",
+        },
+    },
+    production: {
```
daniel0ar, line 13: I see these two have the same params, is this intended?

apps/copilot-api/routes/auth.ts
```diff
@@ -0,0 +1,112 @@
...
+router.post('/login', async (req: Request, res: Response) => {
+  const {signature, walletAddress, challengeMessage} = req.body;
+
+  if (!isAddress(walletAddress)) {
+    res.status(403).json({error: "Invalid wallet address"})
+    return
+  }
+
+  try {
+    const address = await verifyMessage({
```
daniel0ar, line 29: We could use:  [verifySiweMessage](https://viem.sh/docs/siwe/actions/verifySiweMessage). We would need a public client somewhere in a config file though. Like:
```ts
// config/client.ts
import { createPublicClient, http } from 'viem'
import { mainnet } from 'viem/chains'
 
export const publicClient = createPublicClient({
  chain: mainnet,
  transport: http() // Add an RPC URL
})
```
This is a more standard way of logging in to a web3 app like ours. Since the user has already connected their wallet to the extension a walletClient is available to generate a SIWE message on the front and send both signature and challengeMessage to this endpoint. Need feedback from @SolutionsEngineer, @lennardevertz, or @geoist1 

Also the name `address` does not match the boolean return type. I would use `valid` instead.

apps/copilot-api/routes/auth.ts
```diff
@@ -0,0 +1,112 @@
...
+      message: challengeMessage,
+      address: walletAddress,
+      signature
+    });
+
+    if (!address) {
```
daniel0ar, line 35: Same here, I would use `valid` instead

apps/copilot-api/routes/auth.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const token = jwt.sign(payload, process.env.JWT_SECRET as string, {expiresIn: '7d'})
+
+    res.status(200).send({token})
+  } catch (err) {
+
+  }
+})
+
+router.post('/wallet-address', async (req, res) => {
```
daniel0ar, line 74: If SIWE was to be implemented, this would not be needed, since challenge messagae would be generated on the client side with [createSIWEMessage](https://viem.sh/docs/siwe/utilities/createSiweMessage)

apps/copilot-api/routes/auth.ts
SolutionsEngineer: yeah, I think @lennardevertz, suggested that will be better to use SIWE standard

apps/copilot-api/entities/subscribtions.entity.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    @PrimaryColumn({type: 'text', name: 'address'})
+    readonly address!: string;
```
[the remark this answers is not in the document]
lennardevertz: yep it is expected

apps/copilot-api/knexfile.js
MaciejKutynia: I remove it

apps/copilot-api/knexfile.js
MaciejKutynia: for now yes

apps/copilot-api/routes/auth.ts
MaciejKutynia: I added verifySiweMessage

apps/copilot-api/routes/auth.ts
MaciejKutynia: fixed

apps/copilot-api/routes/auth.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      const token = jwt.sign(payload, process.env.JWT_SECRET as string, {expiresIn: '7d'})
+
+      res.status(200).send({token})
+      return;
+    }
+
+    const payload = {
+      user: {
+        id: user.uuid,
```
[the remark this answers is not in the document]
MaciejKutynia: I created user if in future wil we have couple addresses for one user

apps/copilot-api/db.ts
```diff
@@ -0,0 +1,17 @@
+import {DataSource} from 'typeorm'
+import dotenv from "dotenv";
+
+dotenv.config()
```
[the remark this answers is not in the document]
MaciejKutynia: I added env variables to turbo.json, I think now this is not neccesery

apps/copilot-api/routes/auth.ts
daniel0ar: Reverted. This will be added in a separate PR from me.

PR #204, opened 2024-12-28
Copilot siwe

MateuszStawski:


daniel0ar (approved):
Correctly implements dynamically forming the SIWE message by receiving a domain parameter in the API. Changed the type in the extension side and sent it in the hook handle function.

lennardevertz:
please siwtch targeted branch to copilot-dev

lennardevertz:
> Correctly implements dynamically forming the SIWE message by receiving a domain parameter in the API. Changed the type in the extension side and sent it in the hook handle function.

Is this correctly implemented from the backend api side already? is verification successful?

daniel0ar:
As it is right now, the API won't respond with a jwt token unless verification is succesful. In my tests, all signed messages sent by the user are being verified succesfully with the provided publicClient on the API `/login` endpoint.

The only thing I would add for security is the nonce verification too, like this (line 29 of `copilot-api/routes/auth.ts`):
```ts
const valid = await publicClient.verifySiweMessage({
      address: walletAddress,
      message,
      signature,
      nonce,
    });
```

and line 21 of the same file:
```ts
const { signature, walletAddress, message, nonce } = req.body;
```

This means that the user on the extension side should also send the nonce they received from `/wallet-address` API endpoint.

This can be a separate PR or it can be included here.

lennardevertz:
> As it is right now, the API won't respond with a jwt token unless verification is succesful. In my tests, all signed messages sent by the user are being verified succesfully with the provided publicClient on the API `/login` endpoint.
> 
> The only thing I would add for security is the nonce verification too, like this (line 29 of `copilot-api/routes/auth.ts`):
> 
> ```ts
> const valid = await publicClient.verifySiweMessage({
>       address: walletAddress,
>       message,
>       signature,
>       nonce,
>     });
> ```
> 
> and line 21 of the same file:
> 
> ```ts
> const { signature, walletAddress, message, nonce } = req.body;
> ```
> 
> This means that the user on the extension side should also send the nonce they received from `/wallet-address` API endpoint.
> 
> This can be a separate PR or it can be included here.

lets do separate

PR #244, opened 2025-01-09
feat: preload name and pfp before displaying toast

MateuszStawski:
Added preload of ensName and ensAvatar before displaying toast. 
Fixed disabled state for BUY button in modal.

### DIFF
--- a/apps/extension/src/shared/utils/wallet-utils.ts
+++ b/apps/extension/src/shared/utils/wallet-utils.ts
@@ -4,10 +4,3 @@
 export const getShortWalletHex = (wallet: string) => {
   return `${wallet.slice(0, 6)}...${wallet.slice(-4)}`;
 };
-
-/**
- * Returns if text is wallet address
- */
-export const isWalletHex = (text: string) => {
-  return text.startsWith('0x');
-};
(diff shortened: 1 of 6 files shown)


apps/extension/src/shared/utils/wallet-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+export const getShortWalletHex = (wallet: string) => {
+  return `${wallet.slice(0, 6)}...${wallet.slice(-4)}`;
+};
+
+/**
+ * Returns if text is wallet address
+ */
+export const isWalletHex = (text: string) => {
+  return text.startsWith('0x');
```
21142, line 12: I think we should use viem's utility for this

```
import { isAddress } from 'viem';

/**
 * Returns true if the text is a valid wallet address.
 */
export const isWalletHex = (text: string): boolean => {
  return isAddress(text);
};
```

MateuszStawski: I just removed isWalletHex from wallet-utils and used isAddress to don't have duplicated function which only redirect string to isAddress.

21142 (approved):
lgtm now, thanks

PR #272, opened 2025-01-11
fix: Temporarily remove api key

daniel0ar:
## Overview
Commented line  94 of apps/copilot-api/index.ts so that requests to the Li.Fi API are not hitting limits now (until repeated requests to /get-quote are fixed).


lennardevertz:
closing as its not needed anymore

PR #282, opened 2025-01-14
add API docs with endpoints

MaciejKutynia:


lennardevertz:
please add some mention of the swapData format users can expect from subscribing

PR #288, opened 2025-01-14
fix error with missing icon

MaciejKutynia:


apps/extension/src/final/notifications-popup/utils/token-icon.tsx
```diff
@@ -12,12 +12,15 @@ export const TokenIcon: React.FC<TokenIconProperties> = ({
 }) => {
   const [icon, setIcon] = useState<JSX.Element | null>(null);
 
+  const tokenImageArray = tokenImage.split('/');
```
daniel0ar, line 15: Why did you decide to go for split instead of something like tokenImage.startsWith?

apps/extension/src/final/notifications-popup/utils/token-icon.tsx
```diff
@@ -12,12 +12,15 @@ export const TokenIcon: React.FC<TokenIconProperties> = ({
...
+  const isCorrectImage = tokenImageArray[0] === 'data:image';
+
   useEffect(() => {
     const loadIcon = () => {
       if (tokenImage === 'IdrissToken') {
         setIcon(<Icon name="IdrissToken" size={24} className="size-6" />);
       } else {
-        if (tokenData && tokenImage) {
+        if (tokenData && isCorrectImage) {
```
daniel0ar, line 23: Will this checkwork with tokenImage=""? Or tokenImage null

MaciejKutynia: It checks if image have data:image, all other options are false

apps/extension/src/final/notifications-popup/utils/token-icon.tsx
MaciejKutynia: I didn't think of that :D

daniel0ar: Ok no worries I was just curious 😄

PR #290, opened 2025-01-14
Refactor: Use new copilot token list from LiFi

daniel0ar:
## Overview
Previous token list was too short on tokens (only had 50). This new one includes more than 800 tokens.

## How it was solved
Used the new endpoint `https://li.quest/v1/tokens?chains=bas` (notice the filter with Base chain "chains=bas")

## How to test it
Execute any trade on a subscribed address on Jumper exchange with any Base token, it should reflect and see token icons on the IDRISS copilot notification. I tested $MISSER (token address `0x0718F45Bbf4781cE891e4e18182F025725F0fc95`)

### DIFF
--- a/apps/extension/src/application/trading-copilot/commands/get-tokens-list.ts
+++ b/apps/extension/src/application/trading-copilot/commands/get-tokens-list.ts
@@ -7,7 +7,9 @@ import {
 } from 'shared/messaging';
 
 type Response = {
-  items: Record<string, string>[];
+  tokens: {
+    [chainId: string]: Record<string, string>[]; // TODO: Correctly type record depending on API response
+  }
 };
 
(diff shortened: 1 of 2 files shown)


daniel0ar:
There is a comment for a TODO, to correctly type the api response token item. I did not do it because of urgency in deployment but it could be something like:
```ts
type Token = {
  chainId: number;
  address: string;
  symbol: string;
  name: string;
  decimals: number;
  priceUSD: string;
  coinKey: string;
  logoURI: string;
}

type Response = {
  tokens: {
    [chainId: string]: Token[];
};
```
This presents some refactoring on `notifications-popup.tsx` and would mean that using a different token list also requires a refactor on these types.

PR #322, opened 2025-01-20
fix: Linter warnings on donation-notification and use-geo-location

daniel0ar:
## Overview

Removed these warnings from running linter:
```
@idriss-xyz/main-landing:lint:fix: ./src/app/creators/obs/components/donation-notification.tsx
@idriss-xyz/main-landing:lint:fix: 49:7  Warning: Classname 'bg-opacity-80' should be replaced by an opacity suffix (eg. '/80')  tailwindcss/migration-from-tailwind-2
@idriss-xyz/main-landing:lint:fix: 54:9  Warning: Using `<img>` could result in slower LCP and higher bandwidth. Consider using `<Image />` from `next/image` to automatically optimize images. This may incur additional usage or cost from your provider. See: https://nextjs.org/docs/messages/no-img-element  @next/next/no-img-element
@idriss-xyz/main-landing:lint:fix: 61:26  Warning: Classname 'text-gray-900' is not a Tailwind CSS class!  tailwindcss/no-custom-classname
@idriss-xyz/main-landing:lint:fix: 65:27  Warning: Classname 'text-gray-900' is not a Tailwind CSS class!  tailwindcss/no-custom-classname
@idriss-xyz/main-landing:lint:fix: 
@idriss-xyz/main-landing:lint:fix: ./src/components/token-section/hooks/use-geo-location.ts
@idriss-xyz/main-landing:lint:fix: 112:18  Warning: The ref value 'controller.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'controller.current' to a variable inside the effect, and use that variable in the cleanup function.  react-hooks/exhaustive-deps
```

### DIFF
--- a/apps/main-landing/.eslintrc.cjs
+++ b/apps/main-landing/.eslintrc.cjs
@@ -17,9 +17,11 @@ module.exports = {
         '**/.*.mjs',
         '**/*.d.ts',
         'src/app/**/*',
+        'src/components/**/*',
       ],
       rules: {
         'import/no-default-export': 'off',
+        '@next/next/no-img-element': 'off',
       },
     },
   ],
--- a/apps/main-landing/next.config.ts
+++ b/apps/main-landing/next.config.ts
@@ -167,5 +167,4 @@ const nextConfig: NextConfig = {
   },
 };
 
-// eslint-disable-next-line import/no-default-export
 export default withBundleAnalyzer({ enabled: false })(nextConfig);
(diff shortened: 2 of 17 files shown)


21142:
My only question is whether we want to use Next.js's `Image` component instead of a regular ``. I recall that @SolutionsEngineer submitted [PR#198](https://github.com/idriss-xyz/core/pull/198) a while ago, where all `Image` components were replaced in favor of regular `` elements. If that's the case, we might want to suppress this warning from Next.js and stick with using ``.

Other than that, everything looks great! Thanks for cleaning this up. There's no need to check the UI of the donation notification afterward, as my next PR will align its design with the copilot's toast notification.

21142:
If we decide not to use Next.js's `Image` component across the repository (or at least for the `main-landing` part), we should disable this rule in the ESLint configuration.

image

Doing so will suppress all related warnings and ensure consistency throughout the codebase.

PR #327, opened 2025-01-22
Api refactor

MaciejKutynia:


apps/copilot-api/services/auth.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+  };
+
+  return jwt.sign(payload, process.env.JWT_SECRET as string, {
+    expiresIn: '60d',
+  });
+}
+
+export async function validateMessage(
+  walletAddress: `0x${string}`,
```
lennardevertz, line 35: Lets use `Hex` imported from viem instead of `0x${string}` https://viem.sh/docs/glossary/types#hex

apps/copilot-api/services/auth.ts
```diff
@@ -0,0 +1,84 @@
...
+import { dataSource } from '../db';
+import { SubscribersEntity } from '../entities/subscribers.entity';
+import { createSiweMessage, generateSiweNonce } from 'viem/siwe';
+import { ExpiringMap } from '../utils/nonceMap';
+import { publicClient } from '../config/publicClient';
+import type { Request } from 'express';
+
+const subscribersRepo = dataSource.getRepository(SubscribersEntity);
+
+const expiryTime = 5 * 60 * 1000;
```
lennardevertz, line 11: would it be worth adding a constants file?

apps/copilot-api/services/index.ts
```diff
@@ -0,0 +1,73 @@
...
+import { SubscriptionsEntity } from '../entities/subscribtions.entity';
+import {
+  GetQuoteDataInterface,
+  GetQuoteDataResponseInterface,
+  TopAddressesResponseInterface,
+} from '../types';
+
+const subscriptionsRepo = dataSource.getRepository(SubscriptionsEntity);
+
+export const testSwapData = {
```
lennardevertz, line 12: also move to constants file?

lennardevertz (changes requested):

apps/copilot-api/services/index.ts
MaciejKutynia: moved

apps/copilot-api/services/auth.ts
MaciejKutynia: moved

apps/copilot-api/services/auth.ts
MaciejKutynia: changed all `0x${string}` to Hex from viem

PR #352, opened 2025-01-27
feat: check if user meet requirements to be premium user of extension

MateuszStawski:
Added verification if user meet requirements to be premium user, and if user have at least 10000 idriss token can subscribe unlimited ammount of wallets.

Moved logic of checking if free user can subscribe from specific files to useSubscriptions hook.

Improved handling of error state in trading-copilot form.


apps/extension/src/application/trading-copilot/commands/get-ens-staked-balance.ts
```diff
@@ -0,0 +1,53 @@
+import { createPublicClient, formatEther, http } from 'viem';
```
lennardevertz, line 1: Name `get-ens-staked-balance` is misleading, has nothing to do with `ens`. Let's change to simple `get-staked-balance`

apps/extension/src/application/trading-copilot/commands/get-ens-staked-bonus-balance.ts
```diff
@@ -0,0 +1,61 @@
+import {
```
lennardevertz, line 1: same comment with the name as above, remove `ens`

apps/extension/src/application/trading-copilot/hooks/use-subscriptions.ts
```diff
@@ -40,6 +44,27 @@ export const useSubscriptions = ({ wallet, addTabListener }: Properties) => {
...
+      address: wallet?.account,
+    }),
+    staleTime: Number.POSITIVE_INFINITY,
+    retryDelay: 5000,
+  });
+  const stakedBalanceSum =
+    Number(stakedBalanceQuery.data) + Number(stakedBonusBalanceQuery.data);
+
+  const isPremiumUser = stakedBalanceSum >= PREMIUM_THRESHOLD;
+  const canSubscribe =
```
lennardevertz, line 65: Should we store this in the extension's storage for some time? If not, how often is this hook executed? Only if a subscription is made?

lennardevertz (approved):

lennardevertz:
Also please merge copilot-dev into your fork, as we are not tracking `claimed-events.json` anymore

apps/extension/src/application/trading-copilot/hooks/use-subscriptions.ts
MateuszStawski: It's fetched everytime when we call useSubscriptions hook, and then we can get this as parameter from destructuring.
image

I don't think we need to store it. We can think about refreshing if needed.

PR #360, opened 2025-01-28
feat: Add fetch claim Events and store on file

daniel0ar:
## Overview
Indexes event data on Claim contract for CLaimed events, and saves new records to a json file on the server called claimEvents. Event is:
`event Claimed(address by, address token, address to, uint256 total, uint256[] claimIndices, bytes memo, bool bonus)`
Data used for json file:
`const { to, total, bonus }`
Structure of json file:
`{address: {bonus: true/false, total: amount}}`

If server restart, it will read events from initial block `25614734`

## Pending
- [ ] Add endpoint to API to return claimed events for address only?
- [ ] Read from frontend and show total (parse units too)
- [ ] Merge copilot-dev to master?

### DIFF
--- a/apps/copilot-api/utils/claimEvents.ts
+++ b/apps/copilot-api/utils/claimEvents.ts
@@ -1,0 +1,57 @@
+import { createPublicClient, http, parseAbiItem } from 'viem';
+import { base } from 'viem/chains';
+import fs from 'fs';
+
+const CLAIMED_EVENT = 'event Claimed(address by, address token, address to, uint256 total, uint256[] claimIndices, bytes memo, bool bonus)';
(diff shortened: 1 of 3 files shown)


apps/copilot-api/utils/claimEvents.ts
```diff
@@ -0,0 +1,57 @@
...
+  try {
+
+    const client = createPublicClient({
+      chain: base,
+      transport: http('https://base-rpc.publicnode.com'),
+    });
+
+    const parsedEvent = parseAbiItem(CLAIMED_EVENT);
+
+    const claimLogs = await client.getLogs({
```
lennardevertz, line 20: we probably have to figure out iterating like in creators, this block range will be too big

PR #389, opened 2025-02-04
- add creator link logger api

lennardevertz:


apps/main-landing/data/creator-init-links.json
SolutionsEngineer, line 1: just to make sure, do we want to add this to the .gitignore?

turbo.json
```diff
@@ -15,7 +15,8 @@
         "RAILWAY_PUBLIC_DOMAIN",
         "DEV_LOGIN_PASSWORD",
         "PUBLIC_ACCESS_ENABLED",
-        "CI"
+        "CI",
+        "TOP_ADDRESSES_SECRET"
```
SolutionsEngineer, line 19: as we are using this env in the main-landing I think we should add it to the .env.example and the environment.d.ts

apps/main-landing/data/creator-init-links.json
lennardevertz: thanks for checking, this will not be a file that changes, so all good

lennardevertz:
Good to merge @SolutionsEngineer?

SolutionsEngineer (approved):

PR #394, opened 2025-02-09
- fix event parsing for solana swaps on jupiter and raydium

lennardevertz:


apps/copilot-api/utils/swapDataParsers.ts
```diff
@@ -112,48 +137,72 @@ export async function parseJupiterSwap(
...
+        symbol: null,
+        name: null,
+        logoURI: null,
+        network: 'SOLANA',
+      };
+    }
+    if (transfer.toUserAccount === feePayer) {
+      tokenIn = {
+        address: transfer.mint,
+        amount: transfer.tokenAmount,
```
SolutionsEngineer, line 177: just to make sure, in the tokenOut we have Math.abs(...), should we also apply it in here?

apps/copilot-api/utils/swapDataParsers.ts
```diff
@@ -112,48 +137,72 @@ export async function parseJupiterSwap(
...
-    symbol: null,
-    name: null,
-    logoURI: null,
-    network: 'SOLANA',
-  };
+  for (const transfer of nativeTransfers) {
+    if (transfer.fromUserAccount === feePayer && !tokenOut) {
+      tokenOut = {
+        address: 'So11111111111111111111111111111111111111112',
+        amount: transfer.amount / 1e9,
```
SolutionsEngineer, line 191: just like above, most likely it's intentional, but do we want to add Math.abs here?

apps/copilot-api/utils/swapDataParsers.ts
lennardevertz: its fine without as tokenIn will always be positive

PR #406, opened 2025-02-16
feat: share reusable utils

MateuszStawski:
Moved reusable utils from extension and main-landing catalogs to packages/constants. Removed also outdated types like custom Hex.


apps/extension/src/application/agora/utils.ts
```diff
@@ -1,4 +1,6 @@
-import { CHAIN, getBlockExplorerUrl } from 'shared/web3';
+import { getDefaultBlockExplorerUrl } from '@idriss-xyz/constants';
+
+import { CHAIN } from 'shared/web3';
```
SolutionsEngineer, line 3: why we are leaving the `CHAIN` in the `extension/shared/web3`?

apps/extension/src/final/notifications-popup/components/trading-copilot-dialog/trading-copilot-dialog.tsx
```diff
@@ -296,9 +304,35 @@ const TradingCopilotDialogContent = ({
               </span>
             </span>
           </p>
-          <p className="text-body6 text-mint-700">
-            <TimeDifferenceCounter timestamp={dialog.timestamp} text="ago" />
-          </p>
+          <TooltipProvider delayDuration={400}>
```
SolutionsEngineer, line 307: What do you think about extracting this to a separate component, so it can be used here and in `trading-copilot-toast`?

apps/extension/src/shared/web3/constants.ts
SolutionsEngineer, line 1: what do you think about moving all this to the `packages/contants`?

apps/extension/src/shared/web3/utils.ts
SolutionsEngineer, line 1: what do you think about moving this to the `packages/utils`?

apps/main-landing/src/app/creators/donate/constants.ts
SolutionsEngineer, line 1: what do you think about moving this to the `packages/constants`?

apps/main-landing/.env.example
```diff
@@ -2,4 +2,5 @@ RAILWAY_PUBLIC_DOMAIN=
 DEV_LOGIN_PASSWORD=
 PUBLIC_ACCESS_ENABLED=
 CREATOR_LINKS_SECRET=
-CREATOR_LINKS_SECRET=
+ZAPPER_API_KEY=
+DATABASE_URL=
```
SolutionsEngineer, line 6: do we need DATABASE_URL in the `main-landing` app?

apps/main-landing/package.json
```diff
@@ -32,6 +32,7 @@
     "lodash": "^4.17.21",
     "mipd": "^0.0.7",
     "next": "15.0.1",
+    "pg": "^8.13.1",
```
SolutionsEngineer, line 35: do we `pg` in the `main-landing` app?

packages/constants/src/utils.ts
SolutionsEngineer, line 1: what do you think about keeping the separation and moving the utils to `packages/utils`?

turbo.json
```diff
@@ -29,9 +29,11 @@
         "ENVIRONMENT",
         "RAILWAY_PUBLIC_DOMAIN",
         "DEV_LOGIN_PASSWORD",
+        "ZAPPER_API_KEY",
         "PUBLIC_ACCESS_ENABLED",
         "CI",
-        "CREATOR_LINKS_SECRET"
+        "CREATOR_LINKS_SECRET",
+        "DATABASE_URL"
```
SolutionsEngineer, line 36: is `DATABASE_URL` needed in the `main-landing` app?

SolutionsEngineer:
I've left a few comments, but in general, it looks good. Great job! It will definitely be much easier to maintain everything.

apps/extension/src/application/agora/utils.ts
MateuszStawski: Because CHAIN from shared have less chains than extension, and in shared/web3 we define extended list of CHAINS for extension

apps/extension/src/shared/web3/constants.ts
MateuszStawski: I think we shouldn't move all utils and constants from all catalogs to shared/constants because in that way we would have probably 5x more code in shared/constants while most of it isn't used in both apps (main-landing and extension) but only in one of them - which totally misses the point of having things in shared/constants.

apps/extension/src/final/notifications-popup/components/trading-copilot-dialog/trading-copilot-dialog.tsx
MateuszStawski: You talking about Tooltip? We have currently it as component in packages/ui/src/components. Is there anything else you would like to suggest?

apps/main-landing/src/app/creators/donate/constants.ts
MateuszStawski: Let's continue this topic here: https://github.com/idriss-xyz/core/pull/406#discussion_r1966529745

apps/extension/src/shared/web3/utils.ts
MateuszStawski: Let's continue this topic here: https://github.com/idriss-xyz/core/pull/406#discussion_r1966529745

packages/constants/src/utils.ts
MateuszStawski: Yeah, that's great idea. I will work tomorrow on that.

turbo.json
MateuszStawski: Yeah, we can remove it as we moved our DB and Zapper calls to apps/creator-api.

apps/main-landing/.env.example
MateuszStawski: Yeah, we can remove it as we moved our DB and Zapper calls to apps/creator-api.

apps/main-landing/package.json
MateuszStawski: I think we dont, this must be leftover from draft stage.

packages/constants/src/utils.ts
MateuszStawski: Moved.

apps/extension/src/final/notifications-popup/components/trading-copilot-dialog/trading-copilot-dialog.tsx
SolutionsEngineer: I was thinking about creating a small component with this tooltip implementation
```
<TooltipProvider delayDuration={400}>
  <Tooltip>
    <TooltipTrigger asChild>
      <p className="w-fit text-body6 text-mint-700">
        {getTimeDifferenceString({
          text: "ago",
          variant: "short",
          timestamp: dialog.timestamp,
        })}
      </p>
    </TooltipTrigger>

    <TooltipContent className="z-portal w-fit bg-black text-white">
      <p className="text-body6">
        {new Intl.DateTimeFormat("en-US", {
          year: "numeric",
          month: "2-digit",
          day: "2-digit",
          hour: "2-digit",
          minute: "2-digit",
          second: "2-digit",
          hour12: false,
        })
          .format(new Date(dialog.timestamp))
          .replaceAll("/", "-")}
      </p>
    </TooltipContent>
  </Tooltip>
</TooltipProvider>;
```

apps/extension/src/shared/web3/constants.ts
SolutionsEngineer: I'm wondering if, in the long run, we should remove shared/constants. I'm also fine revisiting this later after we discuss how much we want to differentiate the apps and whether we should add something to the packages app immediately or only when it's used in multiple apps. However, in that case, we'd need to ensure it's not already defined in one of the apps before adding a new constant. 
Thanks for the feedback on this, it's a challenging topic to define the right flow and boundaries for the monorepo

apps/extension/src/final/notifications-popup/components/trading-copilot-dialog/trading-copilot-dialog.tsx
MateuszStawski: Im not sure if it's worth of doing, because we use it twice in extension and once in main-landing, so creating it only for 2 ussages in main-landing is kinda useless - while creating it as shared component for 3 ussages in main-landing and extension is even more excessive.

lennardevertz: @MateuszStawski is this something you would want to tackle in this PR?

MateuszStawski: These tooltips are already component from shared/ui, if i understand @SolutionsEngineer correctly he was thinking about doing from this Tooltip implementation new component with specific content in Trigger and Content which is kinda useless in my opinion, because we use it only three times. I think we can resolve this conversation and merge this PR.

SolutionsEngineer: I see value in creating a separate component for this implementation, as it would make the code cleaner and easier to maintain in case of style or format changes. But since it’s only used in a few places, is fairly small, and doesn’t have much logic, I’m also fine with keeping it as it is

SolutionsEngineer (approved):

PR #424, opened 2025-02-24
feat: reopen airdrop

MateuszStawski:
Reopened aidrop claim flow
image

### DIFF
--- a/apps/main-landing/src/app/claim/components/letter/letter-content.tsx
+++ b/apps/main-landing/src/app/claim/components/letter/letter-content.tsx
@@ -15,8 +15,9 @@ import geoist_avatar from './assets/geoist_avatar.png';
 import levertz_avatar from './assets/levertz_avatar.png';
 import user_avatar from './assets/user_avatar.png';
 
+const AIRDROP_ENABLED = true;
+
 export const LetterContent = () => {
-  const airdropEnabled = true;
   const { setCurrentContent } = useClaimPage();
 
   return (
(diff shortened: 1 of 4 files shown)


apps/main-landing/src/app/claim/components/letter/letter-content.tsx
```diff
@@ -6,12 +6,20 @@ import {
...
+import { Button } from '@idriss-xyz/ui/button';
+import { GeoConditionalButton } from '@idriss-xyz/ui/geo-conditional-button';
+
+import { useClaimPage } from '../../claim-page-context';
 
 import geoist_avatar from './assets/geoist_avatar.png';
 import levertz_avatar from './assets/levertz_avatar.png';
 import user_avatar from './assets/user_avatar.png';
 
+const AIRDROP_ENABLED = true;
```
SolutionsEngineer, line 18: should some env be resolved as a value here?

MateuszStawski: I think env isn't place for thing like that, eventually we can move it to constants.ts in claim catalog. I only added it as addition to dont have to remove / add button in code everytime when we want to disable/enable it.

MateuszStawski: @lennardevertz what do you think about that?

SolutionsEngineer: I was thinking about using env variables to allow simple changes without requiring code changes or PRs, but I see your point—sounds good 👌

SolutionsEngineer (dismissed):

apps/main-landing/src/app/claim/components/letter/letter-content.tsx
SolutionsEngineer: Great idea to toggle the state instead of removing and reapplying the code

lennardevertz (approved):

PR #426, opened 2025-02-24
ci: Fix linter incompatible base64Secret type (Buffer)

daniel0ar:
## Overview
Fixed a type-check warning that was popping up locally and did not trigger ci on Github, but always triggered an error on local terminal. The error:
```
Argument of type 'Buffer' is not assignable to parameter of type 'Input'.
  Type 'Buffer' is not assignable to type 'Uint8Array<ArrayBufferLike>'.
    The types of 'slice(...).buffer' are incompatible between these types.
      Type 'ArrayBufferLike' is not assignable to type 'ArrayBuffer'.
        Type 'SharedArrayBuffer' is missing the following properties from type 'ArrayBuffer': resizable, resize, detached, transfer, transferToFixedLength
```

## How it was fixed
Created a new Uint8Array around the base64Secret which was Buffer type.

### DIFF
--- a/apps/extension/src/application/polymarket/constants.ts
+++ b/apps/extension/src/application/polymarket/constants.ts
@@ -25,7 +25,7 @@ export const EMPTY_MARKET_FORM = {
   selectedTokenId: '',
 } as const;
 
-export const EVENT_URL_REGEX = /https:\/\/polymarket\.com\/event\/[^\s"']+/g;
+export const EVENT_URL_REGEX = /https:\/\/poly\.market(?:\.com)?\/[^\s"']+/g;
 
 export const ROUNDING_CONFIG: Record<TickSize, RoundConfig> = {
   '0.1': {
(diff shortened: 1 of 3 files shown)


SolutionsEngineer (dismissed):

apps/extension/src/application/polymarket/constants.ts
```diff
@@ -25,7 +25,7 @@ export const EMPTY_MARKET_FORM = {
   selectedTokenId: '',
 } as const;
 
-export const EVENT_URL_REGEX = /https:\/\/polymarket\.com\/event\/[^\s"']+/g;
+export const EVENT_URL_REGEX = /https:\/\/poly\.market(?:\.com)?\/[^\s"']+/g;
```
SolutionsEngineer, line 28: Just to make sure do we want to make the dot in the middle of domain optional just as the `.com` ?
I mean `...poly\.?market...` to match `poly.market` and `polymarket.com`?

SolutionsEngineer (dismissed):

apps/extension/src/application/polymarket/constants.ts
lennardevertz: very good question, ill adjust it!

SolutionsEngineer (approved):

PR #434, opened 2025-03-06
fix: Prevent multiple onSubmit calls for Farcaster name submissions

lennardevertz:
fix: Await onSubmit calls in Farcaster name handling block


daniel0ar (approved):
Gtg. Tested and working fine.

PR #462, opened 2025-03-18
feat: donation notification enlargement

MateuszStawski:
Updated width of notification to automatically set width between 400 to 430px.

image
image
image

### DIFF
--- a/apps/main-landing/src/app/creators/obs/components/donation-notification.tsx
+++ b/apps/main-landing/src/app/creators/obs/components/donation-notification.tsx
@@ -64,7 +64,7 @@ const DonationNotification = ({
       aria-live="polite"
       nonce={txnHash}
       style={style}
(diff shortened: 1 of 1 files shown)


apps/main-landing/src/app/creators/obs/components/donation-notification.tsx
```diff
@@ -64,7 +64,7 @@ const DonationNotification = ({
       aria-live="polite"
       nonce={txnHash}
       style={style}
-      className={`absolute left-0 top-0 m-3 flex w-[400px] items-start gap-x-2 rounded-xl p-4 shadow-lg transition-opacity duration-1000 ${bgColor} ${
```
lennardevertz, line 67: lets use 430 as max width, but not hardcoded.. so it can be between 400-430px, depending on the name length

lennardevertz (changes requested):

lennardevertz (approved):

PR #489, opened 2025-03-28
- refactor DonationData to decrease api dependency

lennardevertz:
WARNING

Do not run locally until you want to actually migrate your db.


apps/main-landing/src/app/creators/donate/components/history/donate-history-item.tsx
```diff
@@ -161,9 +133,8 @@ export default function DonateHistoryItem({ tip, showReceiver }: Properties) {
 
           <p className="align-middle text-body5 text-neutral-600">
             {tipComment
-              ? tipComment.stringValue !== 'N/A' &&
-                tipComment.stringValue !== ' '
-                ? tipComment.stringValue
+              ? tipComment !== 'N/A' && tipComment !== ' '
```
MateuszStawski, line 136: Can we move this condition to API?

lennardevertz: please re-check

MateuszStawski: Looks ok, but I'm not sure if the API actually returns undefined when comment is N/A or empty string. If you're sure that those 2 cases are properly handled then we can resolve this conversation.

lennardevertz: the api is simply not storing donations that have N/A or empty string

MateuszStawski: good

MateuszStawski (approved):

PR #504, opened 2025-04-08
Feat: Fetch native or token balance on send widget

daniel0ar:
## Overview
Display "Not enough balance" message on idriss-send widget.

## Modifications
- Extract `blockchain-clients.ts` from creators obs to external package. It now lives in @idriss-xyz/blockchain-clients and is shared betwenn main-landing and extension.
- Refactored `GetEnsBalanceCommand` to accept a chainId and use one of the public clients in `blockchain-clients.ts` 
- Added `GetTokenBalanceCommand` to fetch balance from a token contract. Moved ERC_20_ABI to shared/web3 constants file and updated imports. Also uses one of the public clients from `blockchain-clients.ts`
- Modified `use-sender` and `send-widget` to use these commands on extension, fetch balances and return error message on SEND button click (x, farcaster donation widget) saying "Not enough CURRENCY. Add funds first and try again." where CURRENCY is the token symbol or ETH if native.


lennardevertz:
Fix: remove error when user makes an input change on the send widget

lennardevertz (approved):

PR #572, opened 2025-06-18
Feat/privy migration

daniel0ar:
## Overview
Migrate all dynamic login code to the use of Privy instead.


lennardevertz:
closing in favor of #579

PR #594, opened 2025-07-25
QA 71, 72, 33

lennardevertz:


apps/main-landing/src/app/creators/components/oauth-callback-handler.tsx
```diff
@@ -55,6 +53,11 @@ export function OAuthCallbackHandler() {
 
         if (existingCreator) {
           setCreator(existingCreator);
+          if (existingCreator.doneSetup) {
```
daniel0ar, line 56: Nice moved here!

packages/ui/src/components/breadcrumb/breadcrumb-navigation.tsx
```diff
@@ -123,11 +123,7 @@ export function BreadcrumbNavigation({
           return (
             <React.Fragment key={item.path}>
               <BreadcrumbItem>
-                {item.isCurrentPage ? (
-                  <BreadcrumbPage>{item.name}</BreadcrumbPage>
-                ) : (
-                  <BreadcrumbLink href={item.path}>{item.name}</BreadcrumbLink>
-                )}
+                <BreadcrumbPage>{item.name}</BreadcrumbPage>
```
daniel0ar, line 125: I see, we don't care about navigation since we don't really have deep page navigation, level 0 (home) does not exist, level 1 (earnings, setup,..) also does not exist. So we are left with level 2 only and don't need to navigate back to level 1 or 0. 

Maybe add a comment in case we care about navigation in the future "If we want navigation, render                   <BreadcrumbLink href={item.path}>{item.name}</BreadcrumbLink> instead"

apps/main-landing/src/app/creators/app/page.tsx
```diff
@@ -1,24 +1,7 @@
...
-  useEffect(() => {
-    if (creator) {
-      if (creator.doneSetup) {
-        router.replace('/creators/app/earnings/stats');
-      } else {
-        router.replace('/creators/app/setup/payment-methods');
-      }
-    }
-  }, [creator, router]);
-
```
daniel0ar, line 21: I think this deleted section should still stay, in case someone enters creators/app directly it is redirected to the correct screen

apps/main-landing/src/app/creators/utils/navigation.ts
```diff
@@ -27,7 +31,11 @@ export const useStartEarningNavigation = () => {
         return;
       }
       setCreator(fetchedCreator);
-      router.push('/creators/app');
+      if (fetchedCreator.doneSetup) {
```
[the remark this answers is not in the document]
lennardevertz: no, you are right

apps/main-landing/src/app/creators/app/page.tsx
lennardevertz: will check

PR #596, opened 2025-07-26
fix: Redirect back to creators if not authenticated (QA 82)

daniel0ar:
## Overview

Fixes a potential workaround the users had to enter and navigate the creators v2 app without being authenticated.

## How it was solved

- Extracted the logic present in creators/app/layout (to redirect the users if privy session not present) to a new hook called use-redirect-not-authenticated
- Call this new hook in all other inner layouts (earnings, setup, profile and ranking) so in case the user is able to click and navigate to them they are still redirected back to creators/login screen


lennardevertz (approved):
nice!

PR #603, opened 2025-07-28
Referral table, fetching user's twitch followers, join endpoint

SzymonKorda:


apps/creator-api/routes/claim-rewards.ts
```diff
@@ -0,0 +1,146 @@
...
+  const address = req.params.address;
+  if (!address) {
+    console.error('Invalid creator address provided');
+    res.status(400).json({ error: 'Invalid creator address provided' });
+    return;
+  }
+  const hexAddress = address as Hex;
+
+  const creatorRepository = AppDataSource.getRepository(Creator);
+  const creator = await creatorRepository.findOne({
```
lennardevertz, line 31: lets actually use ```  if (!req.user?.id) {
    res.status(401).json({ error: 'Unauthorized' });
    return;
  }

  try {
    const creatorRepository = AppDataSource.getRepository(Creator);
    const creator = await creatorRepository.findOne({
      where: { privyId: req.user.id },
    });``` so we don't have to test a frontend provided address parameter. this auth id is passed through privy directly

apps/creator-api/routes/claim-rewards.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  if (rewardsToClaim === 0) {
+    console.error('No rewards to claim');
+    res.status(400).json({ error: 'No rewards to claim' });
+    return;
+  }
+
+  const chainId = 8453;
+  const tokenAddress = '0x000096630066820566162c94874a776532705231' as Hex;
+  const amount = 2;
```
lennardevertz, line 68: use utility function for the amount calculation like explained in our chat discussion

apps/creator-api/routes/join.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    await referralRepository.save(referral);
+
+    res.status(201).json({
+      message: 'Success',
+    });
+  },
+);
+
+const calculateReward = (followersCount: number) => {
```
lennardevertz, line 103: I guess this is the start of the amount util function (comment above)

apps/creator-api/routes/join.ts
```diff
@@ -0,0 +1,126 @@
+import { Router, Request, Response } from 'express';
```
lennardevertz, line 1: lets potentially move these endpoints? or rename the file? not sure what is being `joined` here

apps/creator-api/routes/join.ts
```diff
@@ -0,0 +1,126 @@
...
+      ? twitchUserFollowersInfo.total
+      : 0;
+
+    const reward = calculateReward(followersCount);
+
+    const referral = new Referral();
+    referral.credited = false;
+    referral.referrer = referrer;
+    referral.referred = referred;
+    referral.reward = reward;
```
lennardevertz, line 92: im not 100% if we want to hardcode the reward here in the db actually. I think we only want to save the follower number so we can be flexible on how big the reward can be in real time

apps/creator-api/routes/referral-history.ts
```diff
@@ -0,0 +1,98 @@
...
+
+interface InvitedStreamersData {
+  name: string;
+  profilePictureUrl: string;
+  numberOfFollowers: number;
+  joinDate: Date;
+  reward: number;
+}
+
+router.get('/:address', verifyToken(), async (req: Request, res: Response) => {
```
lennardevertz, line 26: again also here: lets use authed endpoints and not trust url params. we can get the creator with the id param from the auth

apps/creator-api/routes/referral-history.ts
```diff
@@ -0,0 +1,98 @@
...
+    const streamersData = streamers.map((streamer) => {
+      const referred = referrals.find(
+        (referral) => referral.referred.id === streamer.id,
+      );
+      return {
+        name: streamer.name,
+        profilePictureUrl: streamer.profilePictureUrl,
+        numberOfFollowers: referred?.numberOfFollowers,
+        joinDate: streamer.joinedAt,
+        reward: referred?.reward,
```
lennardevertz, line 82: calculate reward with utility function instead of db entry

packages/constants/src/constants.ts
```diff
@@ -379,7 +379,8 @@ export const DEFAULT_ALLOWED_CHAINS_IDS = [
 export const EMPTY_HEX = '0x';
 
 export const COPILOT_API_URL = 'https://copilot-api.idriss.xyz';
-export const CREATOR_API_URL = 'https://creators-api.idriss.xyz';
+// export const CREATOR_API_URL = 'https://core-production-a116.up.railway.app';
```
lennardevertz, line 382: revert this change

apps/creator-api/routes/claim-rewards.ts
```diff
@@ -0,0 +1,146 @@
...
+
+type Response1 = {
+  price: string;
+};
+
+router.post('/:address', verifyToken(), async (req: Request, res: Response) => {
+  const address = req.params.address;
+  if (!address) {
+    console.error('Invalid creator address provided');
+    res.status(400).json({ error: 'Invalid creator address provided' });
```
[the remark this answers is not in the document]
SzymonKorda: Yes, you're right

apps/creator-api/routes/claim-rewards.ts
SzymonKorda: Is such solution ok?

```
 const creatorRepository = AppDataSource.getRepository(Creator);
  const creator = await creatorRepository.findOne({
    where: { privyId: req.user.id },
  });

  if (!creator) {
    console.error('Creator with given address does not exist');
    res
      .status(400)
      .json({ error: 'Creator with given address does not exist' });
    return;
  }
```

apps/creator-api/routes/join.ts
SzymonKorda: Yes, and it's used for reward dynamic calculating base on followers count

apps/creator-api/routes/claim-rewards.ts
SzymonKorda: Should also this validation be added?
```
if (!req.user?.id) {
      res.status(401).json({ error: 'Unauthorized' });
      return;
    }
```

lennardevertz: both yes

apps/main-landing/src/app/creators/utils/index.ts
```diff
@@ -125,7 +126,7 @@ export const editCreatorProfile = async (
     }
 
     const response = await fetch(`${CREATOR_API_URL}/creator-profile/${name}`, {
-      method: 'PATCH',
```
lennardevertz, line 128: why this change

packages/utils/src/donation-utils.ts
```diff
@@ -2,12 +2,38 @@ import { Hex, parseUnits } from 'viem';
...
   TokenBalance,
 } from '../../constants/src';
 
+import { getTokenPerDollar } from './get-token-per-dollar';
+
+export const calculateDollarsInIdrissToken = async (
+  amount: number,
+  chainId: number,
+) => {
+  const tokenAddress = '0x000096630066820566162c94874a776532705231' as Hex;
```
lennardevertz, line 18: please use import IDRISS_TOKEN_ADDRESS, if needed add it to packages constants

apps/main-landing/src/app/creators/utils/index.ts
SzymonKorda: Testing leftover -> good catch. I changed wrong endpoint for testing purposes.

apps/creator-api/routes/claim-rewards.ts
SzymonKorda: Applied!

packages/utils/src/donation-utils.ts
SzymonKorda: Added to `packages/constants/src/constants.ts`

apps/creator-api/routes/join.ts
SzymonKorda: What do you think about `invite-accept`?

lennardevertz: this is purely on the backend side for the api? that works then

PR #604, opened 2025-07-28
Fix/daniel qas

daniel0ar:
## Overview
Solves QAs:
15
11
54
66
and 73


apps/main-landing/src/app/creators/utils/index.ts
```diff
@@ -38,6 +38,7 @@ export type CreatorProfileResponse = {
   displayName?: string;
   profilePictureUrl?: string;
   email?: string;
+  receiveEmails: boolean;
```
lennardevertz, line 41: needs to be optional, same as email, as we are not pushing this to the public api

lennardevertz: And then in creator-profile.ts, search for
      `const { email, obsUrl, ...publicProfile } = profile;`, and add `receiveEmails`  to the excluded parameters of public apis

apps/creator-api/db/entities/creator.entity.ts
```diff
@@ -47,6 +47,9 @@ export class Creator {
   @Column({ type: 'boolean', name: 'done_setup' })
   doneSetup!: Boolean;
 
+  @Column({ type: 'boolean', name: 'receive_emails' })
+  receiveEmails?: Boolean;
```
lennardevertz, line 51: don't we want the default true here?

apps/main-landing/src/app/creators/app/earnings/stats-and-history/page.tsx
```diff
@@ -281,7 +281,8 @@ export default function EarningsStats() {
                               />
                             </div>
                             <span className="text-sm text-gray-300">
-                              {item.tokenData.symbol}
+                              {item.tokenData.symbol}{' '}
+                              {/* TODO: Change for name */}
```
lennardevertz, line 285: fix todo if you already know what to do?

PR #651, opened 2025-08-13
fixes redirect after logout

lennardevertz:
fix opened login popup after login stemming from a race condition after setting creator to null.

in combination with #638 point 2 should work for all cases

### DIFF
--- a/apps/main-landing/src/app/creators/components/hero-section.tsx
+++ b/apps/main-landing/src/app/creators/components/hero-section.tsx
@@ -49,15 +49,9 @@ export const HeroSection = ({ heroButtonReference }: Properties) => {
   const handleStartEarningClick = () => {
     if (window.innerWidth < 1024) {
       setIsMobileNotSupportedOpen(true);
-
       return;
     }
-
-    if (creator) {
-      void originalHandleStartEarningClick();
-    } else {
-      void originalHandleStartEarningClick();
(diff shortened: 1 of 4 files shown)


apps/main-landing/src/app/creators/hooks/use-logout.ts
```diff
@@ -10,9 +10,14 @@ export const useLogout = () => {
...
     await logout();
+
+    router.replace('/creators');
+
     localStorage.removeItem('twitch_new_user_info');
     localStorage.removeItem('custom-auth-token');
-    setCreator(null);
-    router.push('/creators');
+
+    void Promise.resolve().then(() => {
```
daniel0ar, line 19: Why is this promise resolve needed?

apps/main-landing/src/app/creators/components/hero-section.tsx
```diff
@@ -49,15 +49,9 @@ export const HeroSection = ({ heroButtonReference }: Properties) => {
...
-
       return;
     }
-
-    if (creator) {
-      void originalHandleStartEarningClick();
-    } else {
-      void originalHandleStartEarningClick();
-    }
+    void originalHandleStartEarningClick();
```
daniel0ar, line 54: Got it on my PR as well!

PR #662, opened 2025-08-20
Feat: Display Twitch username on donation alert if donor is registered

daniel0ar:
## Overview
Display Twitch username on donation alert f donor is registered with an IDRISS account (Privy). 

## What's included
- Add a creator-address record in the db when a creator profile is created (on Signup).
- Add new util getUsernameOrAnon and replace resolveEnsName for it.
- Remove the donation-overlay page (should not have been in path /donation-overlay, only exists in /donation-overlay/[slug]) and show not found page instead.
- Display twitch name and defualt pfp on donor stats page (replaces ens name and pfp resolution)


apps/creator-api/routes/creator-address.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    if (existingAddressInCreatorAddress || existingAddressInCreator) {
+      res.status(409).json({
+        error: 'Address already exists for a creator',
+      });
+      return;
+    }
+
+    // Create new creator address
+    const creatorAddress = creatorAddressRepository.create({
```
lennardevertz, line 68: do we want to protect this endpoint with verifyToken()? otherwise people can spam random accounts here right?

apps/creator-api/utils/calculate-stats.ts
```diff
@@ -22,9 +22,39 @@ import {
 import { AppDataSource } from '../db/database';
 import { Creator } from '../db/entities';
 
-export function calculateStatsForDonorAddress(
+async function getCreatorNameOrAnon(address: string): Promise<string> {
```
lennardevertz, line 25: we might want to consider moving this into utils... as currently leaderboards (not globbal on the v2 pages, but the ones on the donation pages) still use the util that makes `vitalik.eth` to `vitalik` and also resolves ENS pfp. going forward, anon should have the default pfp, and ens is considered anon. lets make it so the backend actually makes this check and replaces name and pfp before sending leaderboard data to the frontend, as then we dont have to update the twitch extension again. (default pfp probably is handled by passing no pfp url). I believe the endpoint that returns this leaderboard is `tip-history`.

apps/main-landing/src/app/creators/donate/components/donate-form/donate-form.tsx
```diff
@@ -174,10 +179,53 @@ export const DonateForm = forwardRef<HTMLDivElement, Properties>(
       [sfx],
     );
 
+    // Add a record to creator-address table on db (link
+    // wallet address to creator)
+    const createCreatorAddressLink = useCallback(async () => {
+      if (!walletClient?.account.address) return;
+      await fetch(`${CREATOR_API_URL}/creator-address`, {
```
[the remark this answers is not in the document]
daniel0ar: I have changed it. Why not use the `customAuthToken` from useAuth (our own context)?

apps/creator-api/routes/creator-address.ts
daniel0ar: Done!

apps/creator-api/utils/calculate-stats.ts
daniel0ar: This requires some more work and thought. So I will include it probably on another PR

apps/main-landing/src/app/creators/donate/components/donate-form/donate-form.tsx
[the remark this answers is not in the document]
lennardevertz: maybe that's better, idk... but this is used on all other protected endpoints

apps/main-landing/src/app/creators/donate/components/donate-form/donate-form.tsx
```diff
@@ -174,10 +179,57 @@ export const DonateForm = forwardRef<HTMLDivElement, Properties>(
...
+      if (!walletClient?.account.address) return;
+      const authToken = await getAccessToken();
+      await fetch(`${CREATOR_API_URL}/creator-address`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'Authorization': `Bearer ${authToken}`,
+        },
+        body: JSON.stringify({
+          creatorId: donorCreator?.id,
```
lennardevertz, line 194: I believe we can remove the creator.id here as its retrieved by verifyToken on the backend

apps/creator-api/routes/creator-address.ts
```diff
@@ -0,0 +1,89 @@
...
+    // Validate required fields
+    if (!address || !creatorId) {
+      res.status(400).json({
+        error: 'Address and creatorId are required',
+      });
+      return;
+    }
+
+    // Validate address format (basic hex validation)
+    if (!address.startsWith('0x')) {
```
lennardevertz, line 25: use viem's `getAddress` for verification. `0xabc` is not an address.

daniel0ar: Changed to `isAddress` now (from viem too)

apps/creator-api/routes/creator-address.ts
```diff
@@ -0,0 +1,89 @@
...
+import { getAddress, Hex } from 'viem';
+import { verifyToken } from '../db/middleware/auth.middleware';
+
+const router = Router();
+const creatorAddressRepository = AppDataSource.getRepository(CreatorAddress);
+const creatorRepository = AppDataSource.getRepository(Creator);
+
+router.post('/', verifyToken(), async (req: Request, res: Response) => {
+  try {
+    const { address, creatorId } = req.body;
```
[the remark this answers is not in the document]
daniel0ar: Done!

apps/creator-api/routes/siwe.ts
```diff
@@ -75,7 +75,8 @@ router.post(
         res.status(401).json({ error: 'Bad nonce' });
         return;
       }
-      if (siwe.domain !== req.hostname) {
+      const host = (req.get('host') ?? '').split(':')[0] || req.hostname;
```
daniel0ar, line 78: What is this comparing? "http"? or what do we get on req.get('host') and req.hostname?

apps/creator-api/routes/siwe.ts
```diff
@@ -0,0 +1,132 @@
...
+        res.status(401).json({ error: 'SIWE verify failed' });
+        return;
+      }
+
+      nonceStore.delete(req.user.id);
+
+      const formattedAddress = getAddress(data.address);
+      const creatorId = creator.id;
+
+      const existingAddressInCreator = await creatorRepository.findOne({
```
daniel0ar, line 98: I think this search in Creator table is not needed since `address` will always be the Privy address and not an external one (I added it before bc I was thinking in the future we would have wallet signup with privy, and as discussed on team chat this is not needed)

PR #671, opened 2025-09-01
Fix/daniel qas 4

daniel0ar:
## Overview
Includes solutions to QA points:
125, 126, 127, 133, 134, 135, 136


packages/ui/src/components/icon/custom/token/pengu-token.tsx
```diff
@@ -0,0 +1,85 @@
+import type { SVGProps } from 'react';
+
+export const PenguToken = (properties: SVGProps<SVGSVGElement>) => {
+  return (
+    <svg
+      {...properties}
+      width="60"
```
lennardevertz, line 7: should we use 24x24 here like in other tokens? or doenst matter as long it is `<number>x<number>`?

lennardevertz (approved):

PR #725, opened 2025-10-02
fix nft metadata fetcher and amount display on history items

lennardevertz:
1) fix opensea price estimation
2) fix url for not metadata as explained by the // team (id needs to be hex instead of number when the url has `/tokens/`)
3) remove `1` display when sending 1 not on history items

### DIFF
--- a/apps/creator-api/utils/price-fetchers.ts
+++ b/apps/creator-api/utils/price-fetchers.ts
@@ -336,19 +336,28 @@ export async function fetchNftFloorFromOpensea(
     const rawPrice = offer?.value ?? '0';
     const currency = offer?.currency ?? 'WETH';
 
+    // normalize price per-item if ERC1155 bundle
+    let perUnitRaw = BigInt(rawPrice);
+    const consideration = json.protocol_data?.parameters?.consideration?.[0];
+    const amount = consideration?.startAmount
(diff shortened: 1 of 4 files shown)


apps/creator-api/utils/price-fetchers.ts
```diff
@@ -336,19 +336,28 @@ export async function fetchNftFloorFromOpensea(
     const rawPrice = offer?.value ?? '0';
     const currency = offer?.currency ?? 'WETH';
 
+    // normalize price per-item if ERC1155 bundle
+    let perUnitRaw = BigInt(rawPrice);
+    const consideration = json.protocol_data?.parameters?.consideration?.[0];
```
daniel0ar, line 341: Are we sure consideration is always the first item on the list (index zero)?

apps/main-landing/src/app/creators/utils/index.ts
```diff
@@ -112,6 +112,27 @@ export async function getNftMetadata(
 
   if (!uri) return { name: 'NFT', image: undefined, collectionName };
 
+  /* ---------- fix Parallel “/tokens/” URLs ---------- */
```
daniel0ar, line 115: Tricky workaround, so nice you were able to find it!

